PR #3, opened 2018-05-04
Added new route type and refactored net.Prefix

taktv6:
I'm reimplementing route and routetable entirely. This is intended to break the old route table. More PR's with new routetable implementation will come.

### DIFF
--- a/route/bgp.go
+++ b/route/bgp.go
@@ -1,0 +1,103 @@
+package route
+
+// BGPPath represents a set of BGP path attributes
+type BGPPath struct {
+	PathIdentifier uint32
+	NextHop        uint32
+	LocalPref      uint32
+	ASPath         string
+	ASPathLen      uint16
+	Origin         uint8
+	MED            uint32
+	EBGP           bool
+	BGPIdentifier  uint32
+	Source         uint32
+}
+
+func (r *Route) bgpPathSelection() (best *Path, active []*Path) {
+	// TODO: Implement next hop lookup and compare IGP metrics
+	for _, p := range r.paths {
+		if p.Type != BGPPathType {
+			continue
+		}
+
+		if len(active) == 0 {
+			active = append(active, p)
+			continue
+		}
+
+		if active[0].BGPPath.ecmp(p.BGPPath) {
+			active = append(active, p)
+			if !r.bestPath.BGPPath.better(p.BGPPath) {
+				continue
+			}
+
+			best = p
+			continue
+		}
+
+		if !active[0].BGPPath.betterECMP(p.BGPPath) {
+			continue
+		}
+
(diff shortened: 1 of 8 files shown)


route/path.go
```diff
@@ -0,0 +1,26 @@
+package route
+
+type Path struct {
+	Type       uint8
+	StaticPath *StaticPath
+	BGPPath    *BGPPath
```
cedi: i don't like this. It doesn't scale if there are new Paths.
The path itself is quite different for the different protocols. So creating an common interface is not possible here.
Using empty interface would involve reflection which is also no option.
I will think about this. Maybe i have an idea

route/bgp_path_manager.go
```diff
@@ -0,0 +1,63 @@
...
+	return m
+}
+
+func (m *BGPPathManager) pathExists(p BGPPath) bool {
+	if _, ok := m.paths[p]; !ok {
+		return false
+	}
+
+	return true
+}
```
cedi: i woulud rename this function to `lookup` and return `nil` in case the element is not in the map, and a (pointer to the) object if the key exists.
This would speedup the `AddPath` method, since you don't need to lookup the key twice in the map.

Depending on the used hash function this could have a performance impact in case of 700k calls to `AddPath`

route/bgp_path_manager.go
```diff
@@ -0,0 +1,63 @@
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.pathExists(p) {
+		log.Fatalf("Tried to remove non-existent BGPPath: %v", p)
+		return
+	}
+
+	m.paths[p].usageCount--
+	if m.paths[p].usageCount == 0 {
```
cedi: could be also written as `if !m.paths[p].usageCount`

route/bgp.go
```diff
@@ -0,0 +1,105 @@
...
+
+		if len(active) == 0 {
+			active = append(active, p)
+			best = p
+			continue
+		}
+
+		if active[0].BGPPath.ecmp(p.BGPPath) {
+			active = append(active, p)
+			if !r.bestPath.BGPPath.better(p.BGPPath) {
```
cedi: i suggest to refactor this:
```
if r.bestPath.BGPPath.better(p.BGPPath) {
    best = p
}
continue
```

route/bgp_path_manager.go
taktv6: Done

route/bgp_path_manager.go
taktv6: I don't find that one more clear. It requires more knowledge about the language to guess what exactly is happening here.

PR #5, opened 2018-05-05
Implementing Adj rib in

taktv6:


route/route.go
```diff
@@ -80,7 +80,7 @@ func (r *Route) AddPath(p *Path) {
 	r.paths = append(r.paths, p)
 }
 
-// RemovePath removes path `rm` from route `r`. Returns true if removed path was last one. False otherwise.
+// RemovePath removes path `rm` from route `r`. Returns length of path list after removing path `rm`
```
czerwonk: comment references path name ``rm``, parameter is named ``p``

routingtable/client_manager.go
```diff
@@ -1,21 +1,41 @@
...
 	delete(c.clients, client)
 }
+
+// Clients returns a list of registered clients
+func (c *ClientManager) Clients() (ret []RouteTableClient) {
+	for rtc := range c.clients {
+		ret = append(ret, rtc)
+	}
+
+	return ret
```
czerwonk: i would prefer Clients() returning an empty slice instead of nil when no client is registered.

route/route.go
taktv6: fixed.

routingtable/client_manager.go
taktv6: ack. Fixed.

PR #15, opened 2018-05-19
BGP large community support

czerwonk:
* receiving large communities
* sending large communities
* adding large communities by filter
* filtering for exact large communities

### DIFF
--- a/protocols/bgp/packet/path_attributes.go
+++ b/protocols/bgp/packet/path_attributes.go
@@ -171,7 +171,7 @@ func (pa *PathAttribute) decodeNextHop(buf *bytes.Buffer) error {
 func (pa *PathAttribute) decodeMED(buf *bytes.Buffer) error {
 	med, err := pa.decodeUint32(buf)
 	if err != nil {
-		return fmt.Errorf("Unable to decode local pref: %v", err)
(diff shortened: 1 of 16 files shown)


protocols/bgp/packet/path_attributes.go
```diff
@@ -574,3 +669,16 @@ func isEndOfASSset(asPathPart string) bool {
 func fourBytesToUint32(address [4]byte) uint32 {
 	return uint32(address[0])<<24 + uint32(address[1])<<16 + uint32(address[2])<<8 + uint32(address[3])
 }
+
+func read4BytesAsUin32(buf *bytes.Buffer) (uint32, error) {
```
hikhvar: There is a typo in this function name. It should be read4BytesAsUint32.

To reduce duplicated code we can use this function inside decodeUint32.

czerwonk: Thanks. I will fix that :)

czerwonk: fixed the typo, removed the code duplication and uncluttered the deocdes by passing an string for error string generation to decodeUint32, so we can return instant in decodeXYZ.

hikhvar: Looks nice! :+1:

PR #20, opened 2018-05-28
Compare also next hop

hikhvar:


taktv6:
@czerwonk What do you think about this? Do you see any issue with RFC4271?

czerwonk:
In my opinion it does not harm, since all oligatory conditions of 9.1.2.2. are checked before. As far as I seen the big router vendors also added a few more conditions not specified in 4271. Is this a special case for ECMP routes via route server / iBGP (otherwise the next hop condition would distinguish them)?

PR #33, opened 2018-05-30
Fix: Added missing mandatory LocalPref

hikhvar:


czerwonk:
Since 4271 says SHALL include, i think considering it as mandatory is o.k. for us.

PR #37, opened 2018-05-31
Feature/community support

czerwonk:
This PR depends on #15 

RFC1997 support
* receiving communities
* sending communities

### DIFF
--- a/protocols/bgp/packet/community.go
+++ b/protocols/bgp/packet/community.go
@@ -31,7 +31,7 @@ func ParseCommunityString(s string) (uint32, error) {
 	if err != nil {
 		return 0, err
 	}
-	e2 := uint16(v)
+	e2 := uint32(v)
 
-	return e1<<16 + uint32(e2), nil
+	return e1<<16 + e2, nil
 }
--- a/protocols/bgp/packet/community.go
+++ b/protocols/bgp/packet/community.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 )
 
-func CommunityString(v uint32) string {
+func CommunityStringForUint32(v uint32) string {
 	e1 := v >> 16
 	e2 := v - e1<<16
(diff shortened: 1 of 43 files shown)


protocols/bgp/packet/community.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return 0, err
+	}
+	e1 := uint32(v)
+
+	v, err = strconv.ParseUint(t[1], 10, 16)
+	if err != nil {
+		return 0, err
+	}
+	e2 := uint16(v)
```
hikhvar: IMHO this should be 
```
e2 := uint32(v)
```
since you convert it in the next line. ParseUint(*, 10, 16) make it a uint16.

routingtable/filter/term_condition.go
```diff
@@ -18,7 +19,7 @@ func NewTermCondition(prefixLists []*PrefixList, routeFilters []*RouteFilter) *T
 }
 
 func (f *TermCondition) Matches(p net.Prefix, pa *route.Path) bool {
-	return f.matchesAnyPrefixList(p) || f.machtchesAnyRouteFilter(p)
+	return f.matchesAnyPrefixList(p) || f.machtchesAnyRouteFilter(p) || f.machtchesAnyLageCommunityFilter(pa)
```
hikhvar: Regarding code maintenance I would prefer a TermCondition Type for every TermCondition and introduce a Interface for a TermCondition (prefixList, routeFilters, largeCommunity). This makes it easier to add new TermConditions. However we should test it performance wise.

czerwonk: Fixed the unneccessary cast. For refactoring of Matches(net.Prefix, *route.Path) function I will create a new feature branch in which we benchmark performance.

PR #40, opened 2018-06-02
basic rfc1997 (well known communities) filter, code deduplication

czerwonk:
* added basic filtering for RFC1997 well known communities

For now it is placed in AdjRIBOut. I think after we get the filter implementation with the new FSM done, it should be moved in front of the custom filters


hikhvar:
LGTM

PR #42, opened 2018-06-10
Added tests for large community string parsing

hikhvar:


protocols/bgp/packet/large_community.go
```diff
@@ -2,10 +2,13 @@ package packet
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 )
 
+var largeCommunityRegEx = regexp.MustCompile("^\\(?\\d+,\\d+,\\d+\\)?$")
```
czerwonk: ParseLargeCommunityString is used in all UpdateSenders. I'm a little bit concerned about performance. Can we afford to use regex in such performance critical code sections (thinking on serializing and sending a full table with communities on all routes here)

protocols/bgp/packet/large_community_test.go
```diff
@@ -0,0 +1,86 @@
...
+			err:      &strconv.NumError{Func: "ParseUint", Num: fmt.Sprintf("%d", math.MaxInt64), Err: strconv.ErrRange},
+		},
+		{
+			name:     "to big data part 1",
+			in:       fmt.Sprintf("(1,%d,2)", math.MaxInt64),
+			expected: LargeCommunity{1, 0, 0},
+			err:      &strconv.NumError{Func: "ParseUint", Num: fmt.Sprintf("%d", math.MaxInt64), Err: strconv.ErrRange},
+		},
+		{
+			name:     "to big data part 2",
```
czerwonk: to -> too

protocols/bgp/packet/large_community_test.go
```diff
@@ -0,0 +1,86 @@
...
+			err:      errors.New("malformed large community string [1,2,3]"),
+		},
+		{
+			name:     "missing digit",
+			in:       "(,2,3)",
+			expected: LargeCommunity{},
+			err:      errors.New("malformed large community string (,2,3)"),
+		},
+		{
+			name:     "to big global administrator",
```
czerwonk: to -> too

protocols/bgp/packet/large_community_test.go
```diff
@@ -0,0 +1,86 @@
...
+			in:       fmt.Sprintf("(%d,1,2)", math.MaxInt64),
+			expected: LargeCommunity{},
+			err:      &strconv.NumError{Func: "ParseUint", Num: fmt.Sprintf("%d", math.MaxInt64), Err: strconv.ErrRange},
+		},
+		{
+			name:     "to big data part 1",
```
czerwonk: to -> too

protocols/bgp/packet/large_community.go
czerwonk: @taktv6 what's your opinion here?

hikhvar: IMHO thats a simple regex. Evaluating the regex is O(n) with n is the length of the string. This should be as fast as the split() but gives a more fail save program. I found the internal representation as string rather confusing. If this is performance critical, we should switch from:
```
type BGPPath struct {
...
	Communities      string
	LargeCommunities string
}
```
to
```

type BGPPath struct {
...
	Communities      []packet.Community
	LargeCommunities []packet.LargeCommunity
}
```
and provider the string parsing for users.

czerwonk: I think nobody in the team likes the handling of path attributes as strings. But in the moment it's a tradeoff that allows us to use the path attribute struct itself as key for the map which handles the deduplication. This is really cheap, so until now it was worth it. At this point I think we will get in trouble (staying with this tradeoff) when we want to pass through all unknown attributes. This means every attribute would have to be serialized in string. In my opinion it would cost us a lot more memory than the performance advantage will provide in the future. So I talked to takt today and we will discuss the issue in the next few days.

hikhvar: What map are you referring to? You can use arbitrary structs as map keys. Except for slices. https://blog.golang.org/go-maps-in-action

hikhvar: I run a benchmark:
Old Implementation:
```
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-1-4                5000000               275 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-2-4                5000000               281 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-4-4                5000000               290 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-8-4                5000000               311 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-16-4               5000000               340 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-32-4               5000000               340 ns/op
```
New Implementation:
```
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-1-4                3000000               575 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-2-4                2000000               696 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-4-4                2000000               905 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-8-4                1000000              1373 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-16-4                500000              2143 ns/op
BenchmarkParseLargeCommunityString/BenchmarkParseLargeCommunityString-32-4                500000              3701 ns/op
```
You are correct, this new implementation has a significant performance drop!

PR #44, opened 2018-06-16
Applied filters to FSM

czerwonk:


czerwonk:
Code was picked in #46 because it had to be integrated in the new FSM

PR #53, opened 2018-06-23
Run go vet in CI

q3k:


q3k:
We're dropping this for now as we had some issues with 'go vet'.

PR #63, opened 2018-06-24
Add Route Reflector capablities

BarbarossaTM:


protocols/bgp/packet/path_attributes_test.go
```diff
@@ -1479,6 +1632,106 @@ func TestSerializeCommunities(t *testing.T) {
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(te *testing.T) {
+			buf := bytes.NewBuffer(nil)
+			n := test.input.serializeOriginatorID(buf)
+			if n != test.expectedLen {
+				t.Errorf("Unexpected length for test %q: %d", test.name, n)
+			}
+
+			assert.Equal(t, test.expected, buf.Bytes())
```
czerwonk: pass te instead of t here, otherwise all tests fail when one test case fails. when we start parallelising tests this would hurt us. better rename te to t, then the outer t is hidden

protocols/bgp/packet/path_attributes_test.go
```diff
@@ -1479,6 +1632,106 @@ func TestSerializeCommunities(t *testing.T) {
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(te *testing.T) {
+			buf := bytes.NewBuffer([]byte{})
+			n := test.input.serializeClusterList(buf)
+			if n != test.expectedLen {
+				t.Fatalf("Unexpected length for test %q: %d", test.name, n)
+			}
+
+			assert.Equal(t, test.expected, buf.Bytes())
```
czerwonk: pass te here or rename te to t

routingtable/adjRIBIn/adj_rib_in.go
```diff
@@ -64,6 +68,20 @@ func (a *AdjRIBIn) AddPath(pfx net.Prefix, p *route.Path) error {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
+	// RFC4456 Sect. 8: Ignore route with our RouterID as OriginatorID
+	if p.BGPPath.OriginatorID == 1 {
```
czerwonk: i think this should be a equal comparison against our router id insted of a static 1

routingtable/neighbor.go
```diff
@@ -22,6 +22,12 @@ type Neighbor struct {
 	// Peer is a route server client
 	RouteServerClient bool
 
+	// Peer is a route reflector client
```
czerwonk: Golint demands a comment like: RouteReflectorClient indicates peer is a route reflector client

routingtable/neighbor.go
```diff
@@ -22,6 +22,12 @@ type Neighbor struct {
...
+	RouteReflectorClient bool
+
+	// Our route reflection clusterID
+	ClusterID uint32
```
czerwonk: see obove

PR #67, opened 2018-06-25
add notification when ceasing because of bad peer as (resolves #65)

### LINKED ISSUE #65: Send notification when terminating connection caused by receiving unexpected peer ASN

czerwonk:


q3k:
Merged master into this, this should fix the failing test.

PR #70, opened 2018-06-25
Feature/refactoring bgp path

czerwonk:
* BGPPath not key for maps any more
* BGPPath.ASPath now a slice
* BGPPath.Communities now a slice
* BGPPath.LargeCommunities now a slice


route/bgp_path.go
```diff
@@ -171,16 +172,36 @@ func (b *BGPPath) Prepend(asn uint32, times uint16) {
 		return
 	}
 
-	asnStr := strconv.FormatUint(uint64(asn), 10)
+	first := b.ASPath[0]
```
taktv6: Are we sure this element always exists?

route/bgp_path.go
```diff
@@ -192,6 +213,48 @@ func (p *BGPPath) Copy() *BGPPath {
...
+		b.Source,
+		b.Communities,
+		b.LargeCommunities,
+		b.PathIdentifier)
+
+	r := strings.NewReader(s)
+	h := sha256.New()
+	r.WriteTo(h)
+
+	return fmt.Sprintf("%x", h.Sum(nil))
```
taktv6: Let's pass a byte slice here. Avoids unnecessary allocations.

PR #80, opened 2018-06-26
Feature/sendbuffer plus some refactoring

taktv6:


protocols/bgp/packet/path_attributes_test.go
```diff
@@ -1582,14 +1582,14 @@ func TestSerialize(t *testing.T) {
 func TestSerializeAddPath(t *testing.T) {
 	tests := []struct {
 		name     string
-		msg      *BGPUpdateAddPath
+		msg      *BGPUpdate
 		expected []byte
 		wantFail bool
 	}{
-		{
+		/*{
```
czerwonk: please uncomment or remove

protocols/bgp/server/withdraw_test.go
```diff
@@ -111,7 +111,9 @@ func TestWithDrawPrefixesAddPath(t *testing.T) {
 	}
 	for _, tc := range testcases {
 		buf := bytes.NewBuffer([]byte{})
-		opt := &packet.Options{}
+		opt := &types.Options{
+			AddPathRX: true,
+		}
 		err := withDrawPrefixesAddPath(buf, opt, tc.Prefix, tc.Path)
```
czerwonk: we should consider renaming it to withdrawPrefixes and let this method decide which type of withdraw message is needed internaly based on the options passed

protocols/bgp/types/options.go
```diff
@@ -0,0 +1,7 @@
+package types
+
+// Options represents options to the update sender
```
czerwonk: as well as the decoder and encoder

protocols/bgp/server/withdraw_test.go
czerwonk: nothing to do here for now

protocols/bgp/packet/path_attributes_test.go
taktv6: Fixed.

protocols/bgp/types/options.go
taktv6: Fixed.

PR #82, opened 2018-06-28
New type bio-rt/net/IP to represent IPv4 and IPv6

czerwonk:


net/ip.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+// IPv4FromOctets returns an IPv4 address for the given 4 octets
+func IPv4FromOctets(o1, o2, o3, o4 uint8) IP {
+	return IPv4(uint32(o1)<<24 + uint32(o2)<<16 + uint32(o3)<<8 + uint32(o4))
+}
+
+// IPv6 returns a new `IP` representing an IPv6 address
+func IPv6(higher, lower uint64) IP {
```
BarbarossaTM: IPv6FromString() wäre noch schick *find*

czerwonk: will be added in a further PR, this one was big enough ;).
In the next step the NLRI will be changed, I can introduce a FromString then.

PR #110, opened 2018-07-25
Implement Netlink communication

cedi:
This PR is mainly for reviewing the first draft of the implementation.
I didn't fully tested this yet. I also have no unit-tests till now.

I just wanna hear some opinions :)


cedi:
I'm a little bit unhappy with the current implementation, since i'm throwing all routes i'm learning from the loc_rib directly against the kernel. There is no check if the route was already written, so it could produce a lot of "rtnetlink: file exists" errors.
Doing this a little bit more elegant would mean, that i have to implement a routing-table in between netlink and loc_rib which seems like an massive overkill. It would also requrie me to keep state of the kernel routing table inside the demon which is something i would like to prevent.

I would like to hear some opinions on that from @taktv6 and @czerwonk 

Also how routes are learned from the kernel and written back to the loc_rib did not feels elegant to me.

protocols/system/netlink/netlink.go
```diff
@@ -0,0 +1,224 @@
...
+
+// Add a path to the Kernel using netlink
+// This function is triggered by the loc_rib, cause we are subscribed as
+// client in the loc_rib
+func (n *NetlinkServer) AddPath(pfx net.Prefix, path *route.Path) error {
+	log.WithFields(log.Fields{
+		"Prefix": pfx.String(),
+	}).Debug("AddPath to netlink")
+
+	err := netlink.RouteAdd(n.createRoute(pfx, path))
```
hikhvar: Maybe use https://godoc.org/github.com/vishvananda/netlink#RouteReplace instead of RouteAdd? Replace will do an add if not present, else an replacement

protocols/system/netlink/netlink.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/bio-routing/bio-rd/routingtable"
+	rt "github.com/bio-routing/bio-rd/routingtable"
+	"github.com/bio-routing/bio-rd/routingtable/locRIB"
+	log "github.com/sirupsen/logrus"
+	"github.com/vishvananda/netlink"
+
+	bnet "github.com/bio-routing/bio-rd/net"
+)
+
+type NetlinkServer struct {
```
hikhvar: Isn't NetlinkClient a better name? Or is it a Server in the terminology in the netlink API?

protocols/system/netlink/netlink.go
cedi: Awesome! Thanks

BarbarossaTM: I wouldn't do that. Without checking that the route is from use "proto bio", we might overwrite/replace a route that we did not put into the kernel table, which I would consider a hard bug. So IMO there are some more checks required which is a second strong argument for another bio-internal table which represents the kernel table.

hikhvar: This is true. Could we just reuse https://github.com/bio-routing/bio-rd/blob/master/routingtable/table.go#L12 ? We use this table already in multiple components like adj_rib_out: https://github.com/bio-routing/bio-rd/blob/master/routingtable/adjRIBOut/adj_rib_out.go

cedi:
As we agreed in the call i rework the mentioned points and open a new one.
Of course i will take respect of all comments written here so far

PR #111, opened 2018-07-27
Basic PR for reading global config from files

FabianCernota:
Basic integration of viper. 

Currently reading a global.[yaml, json, toml] from /etc/bio-rd/, $HOME/.bio-rd or ./ )


.travis.yml
```diff
@@ -1,5 +1,5 @@
 dist: trusty
-
+sudo: required
```
hikhvar: Unless we really need sudo, we should not use it. Builds with sudo takes a significant longer time to start.

config/server_test.go
```diff
@@ -78,3 +79,7 @@ func TestAddrIsGreater(t *testing.T) {
 		})
 	}
 }
+
+func TestReadGlobalConfig(t *testing.T) {
```
hikhvar: No tests required?

.travis.yml
```diff
@@ -17,6 +17,8 @@ before_install:
   - wget https://github.com/bazelbuild/bazel/releases/download/0.14.1/bazel-0.14.1-installer-linux-x86_64.sh
   - sha256sum -c .travis.bazel-0.14.1-installer-linux-x86_64.sh.sha256
   - bash bazel-0.14.1-installer-linux-x86_64.sh --user
+  - sudo mkdir /etc/bio-rd
+  - sudo cp global.example.yaml /etc/bio-rd/global.yaml
```
hikhvar: Do we need this directories? ` $HOME/.bio-rd` should be enough.

config/server.go
```diff
@@ -88,8 +112,8 @@ func _getHighestIP(ifs []net.Interface) (net.IP, error) {
 	return nil, fmt.Errorf("No non localhost IPv4 address found on interface lo")
 }
 
-func getLoopbackIP() (net.IP, error) {
-	iface, err := net.InterfaceByName("lo")
+func getLoopbackIP(g *Global) (net.IP, error) {
```
taktv6: Why don't you just pass the interface name (string)?

hikhvar: Or define getLoopbackIP on `Global`

FabianCernota:
@taktv6 @hikhvar Now it should be mergeable :)

config/server.go
```diff
@@ -40,8 +46,41 @@ func (g *Global) SetDefaultGlobalConfigValues() error {
 	return nil
 }
 
-func generateRouterID() (uint32, error) {
-	addr, err := getLoopbackIP()
+func (g *Global) ReadGlobalConfig() error {
+
+	return _readGlobalConfig(g)
```
hikhvar: IMHO just put the content of `_readGlobalConfig` here. Moreover from my point of view `_readGlobalConfig` is a bad name for a Go function since it starts with `_`. That read like python.

https://golang.org/doc/effective_go.html#mixed-caps

hikhvar:
Except function naming LGTM.

Gopkg.toml
```diff
@@ -41,6 +41,14 @@ required = ["github.com/golang/dep", "github.com/q3k/goveralls", "github.com/go-
   branch = "master"
   name = "github.com/taktv6/tflow2"
 
+[[constraint]]
+  name = "github.com/spf13/viper"
+  version = "1.0.2"
+
 [[constraint]]
   branch = "master"
-  name = "github.com/FMNSSun/libhash"
```
taktv6: libhash is necessary for ISIS. Please update this file.

FabianCernota:
@taktv6 Thanks! You can recheck now :)

awlx:
Local tests with bazel fail if there is no local `/home/$user/.bio-rd/global.yaml`, `/etc/bio-rd/global.yaml` or  the corresponding buildpaths. 

```--- FAIL: TestBgpServerPeerSnapshot (0.00s)
    server_test.go:23: server should have started, got err: Failed to load global config: unable to read the global config file: viper.ConfigFileNotFoundError{name:"global", locations:"[/etc/bio-rd /home/$user/.bio-rd /home/$user/.go/src/github.com/bio-routing/bio-rd/protocols/bgp/server]"}
time="2018-10-18T10:23:46+02:00" level=error msg="Unable to serialize BGP Update: general error"
FAIL
 --- FAIL: TestReadGlobalConfig (0.00s)
    server_test.go:90: Error on reading global config file: unable to read the global config file: viper.ConfigFileNotFoundError{name:"global", locations:"[/etc/bio-rd /home/$user/.bio-rd /home/$user/.go/src/github.com/bio-routing/bio-rd/config]"}
FAIL```

File: protocols/bgp/server/server.go [hunk 7]
```diff
@@ -51,6 +51,10 @@ func (b *bgpServer) RouterID() uint32 {
 }
 
 func (b *bgpServer) Start(c *config.Global) error {
+	if err := c.ReadGlobalConfig(); err != nil {
```
taktv6: The BGP server must be independent of the config file in order to be usable as a library.
Config MUST be read outside of protocol (be it BGP, BMP or ISIS) structures and be passed as an argument.

awlx:
There also should be an arg to pass to the binary to override the default paths of the configuration file.

cedi:
Didn't we already agreed on implementing the config part of bio as an gRPC frontent?

(the demon itself is only initiates the protocol servers and wraps all the API function from the bio-library in gPRC functions so for e.g. new BGP-Peers can be added via this gRPC call.
Then it would be easily extendable to implement frontends which can read numerous config files and even implement nice Juniper or Cisco alike CLI's.)

I'm just asking because i didn't had time to attend our bi-weekly video calls in the last time because of some private reasons, and maybe there was consent to now go this way of reading files.
@taktv6 @BarbarossaTM @czerwonk could you please comment on that?
I'm attending our next call for sure to clarify this.

PR #114, opened 2018-09-24
WIP: ISIS packet

taktv6:


protocols/isis/server/interface.go
```diff
@@ -0,0 +1,299 @@
...
+		Version: 1,
+		MaxAreaAddresses: 0,	
+	}
+
+	hdrBuf := bytes.NewBuffer(nil)
+	hdr.Serialize(hdrBuf)
+
+	hdrBuf.Write(helloBuf.Bytes())
+
+	fmt.Printf("Sending Hello: %v\n", hdrBuf.Bytes())
```
czerwonk: should be converted to log output

protocols/isis/server/interface.go
```diff
@@ -0,0 +1,299 @@
...
+	default:
+		log.Errorf("Unknown packet received from %v: %v", src, rawPkt)
+	}
+}
+
+func (n *netIf) processIngressP2PHello(pkt *packet.ISISPacket) {
+	hello := pkt.Body.(*packet.P2PHello)
+	
+	for _, tlv := range hello.TLVs {
+		fmt.Printf("TLV Type: %d\n", tlv.Type())
```
czerwonk: debug output should be converted to logging or removed before merging

protocols/isis/server/interface.go
```diff
@@ -0,0 +1,299 @@
...
+	passive            bool
+	p2p                bool
+	l1                 *level
+	l2                 *level
+	socket             int
+	supportedProtocols []uint8
+}
+
+type level struct {
+	HelloInterval uint16
```
czerwonk: since level is not exported, the attributes should not be exported too, except they would fulfill a interface (which seems not the case here, looks like copy paste of ISISLevelConfig)

protocols/isis/server/interface.go
```diff
@@ -0,0 +1,299 @@
...
+		return fmt.Errorf("failed to send packet: %v", err)
+	}
+
+	return nil
+}
+
+func (n *netIf) p2pHelloTLVs() []packet.TLV {
+
+	l2AdjacencyState, neighborSystemID, neighborExtendedLocalCircuitID := n.p2pL2AdjacencyState()
+	p2pAdjStateTLV := packet.NewP2PAdjacencyStateTLV(l2AdjacencyState, 1234)
```
czerwonk: 1234 seems odd here, is this a valid argument here?

protocols/isis/server/lsdb.go
```diff
@@ -0,0 +1,4 @@
+package server
+
+type lsdb struct {
```
czerwonk: this struct is not used yet, if it is part of already removed code, it should be removed, too

protocols/isis/server/server.go
```diff
@@ -0,0 +1,57 @@
...
+		isis.interfaces[ifs.Name].startReceiver()
+		isis.interfaces[ifs.Name].helloSender()
+	}
+
+	return nil
+}
+
+// Stop stops an ISIS speaker
+func (isis *ISISServer) Stop() error {
+
```
czerwonk: unnecessary empty line

taktv6:
I'm splitting this one into 2.

PR #125, opened 2018-10-17
MP_UNREACH also uses NLRI struct now

czerwonk:


protocols/bgp/packet/mp_unreach_nlri.go
```diff
@@ -4,79 +4,63 @@ import (
...
+	NLRI *NLRI
 }
 
 func (n *MultiProtocolUnreachNLRI) serialize(buf *bytes.Buffer, opt *EncodeOptions) uint16 {
 	tempBuf := bytes.NewBuffer(nil)
 	tempBuf.Write(convert.Uint16Byte(n.AFI))
 	tempBuf.WriteByte(n.SAFI)
-	for _, pfx := range n.Prefixes {
+
+	for cur := n.NLRI; cur != nil; cur = cur.Next {
```
taktv6: Isn't this block redundant to what nlri.go does?

PR #128, opened 2018-10-19
BMP Server

taktv6:


protocols/bgp/server/bmp_router.go
```diff
@@ -0,0 +1,333 @@
...
+	neighborsMu      sync.Mutex
+	logger           *log.Logger
+	runMu            sync.Mutex
+	stop             chan struct{}
+}
+
+type neighbor struct {
+	localAS  uint32
+	peerAS   uint32
+	address  [16]byte
```
BarbarossaTM: assuming the address is the address of the peer -> peerAddress like peerAS?

protocols/bgp/server/bmp_router.go
```diff
@@ -0,0 +1,333 @@
...
+		msg, err := recvBMPMsg(r.con)
+		if err != nil {
+			r.logger.Errorf("Unable to get message: %v", err)
+			return
+		}
+
+		bmpMsg, err := bmppkt.Decode(msg)
+		if err != nil {
+			r.logger.Errorf("Unable to decode BMP message: %v", err)
+			return
```
BarbarossaTM: Killing the server because of a stupid neighbor with broken implementation?

protocols/bgp/server/bmp_router.go
```diff
@@ -0,0 +1,333 @@
...
+
+	if _, ok := r.neighbors[msg.PerPeerHeader.PeerAddress]; ok {
+		return fmt.Errorf("Received peer up notification for %v: Peer exists already", msg.PerPeerHeader.PeerAddress)
+	}
+
+	if len(msg.SentOpenMsg) < packet.MinOpenLen {
+		return fmt.Errorf("Received peer up notification for %v: Invalid sent open message: %v", msg.PerPeerHeader.PeerAddress, msg.SentOpenMsg)
+	}
+
+	sentOpen, err := packet.DecodeOpenMsg(bytes.NewBuffer(msg.SentOpenMsg[19:]))
```
BarbarossaTM: Magic value (19)

protocols/bgp/server/bmp_router.go
```diff
@@ -0,0 +1,333 @@
...
+	if err != nil {
+		return fmt.Errorf("Unable to decode sent open message sent from %v to %v: %v", r.address.String(), msg.PerPeerHeader.PeerAddress, err)
+	}
+
+	if len(msg.ReceivedOpenMsg) < packet.MinOpenLen {
+		return fmt.Errorf("Received peer up notification for %v: Invalid received open message: %v", msg.PerPeerHeader.PeerAddress, msg.ReceivedOpenMsg)
+	}
+
+	recvOpen, err := packet.DecodeOpenMsg(bytes.NewBuffer(msg.ReceivedOpenMsg[19:]))
```
BarbarossaTM: Same

protocols/bgp/server/bmp_router.go
```diff
@@ -0,0 +1,333 @@
...
+	if err != nil {
+		return fmt.Errorf("Unable to decode received open message sent from %v to %v: %v", msg.PerPeerHeader.PeerAddress, r.address.String(), err)
+	}
+
+	addrLen := net.IPv4len
```
BarbarossaTM: As discussed, a comment may be good for this :)

protocols/bgp/server/fsm_address_family.go
```diff
@@ -87,13 +103,11 @@ func (f *fsmAddressFamily) processUpdate(u *packet.BGPUpdate) {
 		return
 	}
 
-	if f.multiProtocol {
```
BarbarossaTM: That's kind of a logic change. Are we sure that IPv4 updates will never code as MP updates, ever?

protocols/bgp/server/bmp_router.go
taktv6: Fixed.

protocols/bgp/server/bmp_router.go
taktv6: We're killing the connection to the stupid neighbour and only the stupid neighbour.

protocols/bgp/server/fsm_address_family.go
taktv6: To be approved by Dan.

czerwonk: If we announce MP cap for IPv4 (which is possible by setting `Peer.AdvertiseIPv4MultiProtocol` to `true` and the other side does so too, we are supposed to send/receive updates for IPv4 as MP_REACH.

czerwonk: That's O.K. for me and the RFC (4760)

PR #129, opened 2018-10-19
Implement Netlink

cedi:
With this commit i'm introducing the netlink protocol to read routes
from the linux kernel routing table to the locRIB.
This also adds the functionality to insert new routes to the linux
kernel routing stack which where learned by other protocol (like e.g.
BGP).

This is done by the protocols/netlink/netlink.go object.
The netlink.go file acts as an facade around the netlink_reader.go and
netlink_writer.go.

netlink_reader itself reads in a defined reding interval new routes from
the kernel and propagates new routes to all subscribed clients.
The netlink reader can also be filtered using the same filter-api as in
adjRibIn/adjRibOut for BGP.
To read new routes to the locRIB you have to register the locRIB to the
netlink_reader using the ClientManager interface, just like you would
register the locRIB to the adjRibIn.

netlink_writer must be subscribed to the locRIB in order to write the
routes from locRIB into the kernel.
This useses the ClientManager interface, just like you would register
the adjRibOut to the locRIB.
netlink_writer itself currently does not support filters, since all
routes form the locRIB should be written to the kernel.
The filter functionality could be easily added here as well.

The netlink_writer currently does not support ecmp. This is subject of
an other commit.

The current implementation could be considered as work in progress, but
the code is mature enough to start a review on it.

PR #131, opened 2018-10-22
Added Makefile for simplicity

cedi:
I as an old fashioned Linux software engineer like Make very much and
found it much easier to just type `make` instead of

    bazel run //:gazelle -- update
    bazel test //...
    bazel coverage //...

everytime i wanted to build.
So i created this Makefile.

It adds the following targets:

    # execute gazelle, test and coverage
    $ make

    # executes `bazel run //:gazelle --update`
    $ make gazelle

    # executes `bazel test //...`
    $ make test

    # executes `bazel coverage //...`
    $ make coverage

    # executes `bazel build //vendor/github.com/golang/dep/cmd/dep` in
    # order to compile dep
    $ make dep

    # removes dep of dep, which breaks gazelle
    $make dep-clean

    # calls `bazel clean`
    $ make clean


cedi:
comment from #129 

> That't true. It's not required for netlink.
> Anyway i would like to get an feeling if the makefile itself considered as an improvement or not.
> I know @taktv6's opinion on Makefiles. But for me as "old" an Linux/Unix engineer typing
> 
>     make
>     make test
>     make gazelle
> 
> is much more convenient than
> 
>     bazel build //:bio-rd
>     bazel test //...
>     bazel run //:gazelle -- update
> 
> Hope i can make my point clear.
> 
> But if the consent is now, that the Makefile is not a desired thing, i will remove it completely.

cedi:
let's close this :joy:

PR #137, opened 2018-11-05
remove bazel

taktv6:


cedi:
i think you missed the README.md
You should update the readme as well, so no one tries to use bazel any longer

PR #175, opened 2019-01-13
since pushd is not available on os x, the script couldn't be run

czerwonk:


taktv6:
This works fine on OS X. It doesn't on Ubuntu 18.04 as /bin/sh is a DASH shell there...

PR #182, opened 2019-01-20
Metrics

czerwonk:


protocols/bgp/server/fsm.go
```diff
@@ -70,6 +70,8 @@ type FSM struct {
 	reason     string
 	active     bool
 
+	counters *fsmConters
```
corny: There is no need to have have pointer here. The struct should be embedded directly.

protocols/bgp/metrics/bgp_address_family_metrics.go
```diff
@@ -0,0 +1,17 @@
+package metrics
+
+// BGPAddressFamilyMetrics provides metrics on AFI/SAFI level for one session
+type BGPAddressFamilyMetrics struct {
```
hikhvar: I miss a metric representing the state of the FSM. 
Currently I compute the following metric:
```
prometheus.NewDesc(
		"peer_states_total",
		"peer states of all connected peers",
		[]string{"peer", "state"},
		nil,
	)
```
However I can not access the required information with the current master since `server.PeerInfo` is gone.

czerwonk: It now exists, without having duplicate states to manage. Thanks for noticing.

PR #184, opened 2019-01-21
Don't check for nil reference in the first line

corny:
Bad style. This should be up to the caller and reduces function calls.


hikhvar:
Actually, I like the current style more. It is a go idiom to make nil a useful value. Moving the nil check outside of the function, we might run into nil pointer errors. 
Yes, there will be less function calls. Do you have any benchmarking numbers how much time we save?

PR #198, opened 2019-03-09
add vrf-id to existing vrf implementation

cedi:
Having IDs to vrf's is crucial for the FIB implementation that follows later, since the VRF-ID decides the routing table in the linux kernel implementation.

Cumulus networks add a feature for "auto assigning" the VRF-id, which i generally like:

```
cumulus@switch:~$ net add vrf rocket vrf-table auto
cumulus@switch:~$ net add interface swp1 vrf rocket
cumulus@switch:~$ net pending
cumulus@switch:~$ net commit
```
https://docs.cumulusnetworks.com/display/DOCS/Virtual+Routing+and+Forwarding+-+VRF

According to the docs the VRF-id is then selected from the range 1001 to 1255
I plan to do another PR where i add this feature as well, to automatically assign the VRF-id based on the information "what is available"

### DIFF
--- a/routingtable/vrf/vrf_registry.go
+++ b/routingtable/vrf/vrf_registry.go
@@ -9,14 +9,16 @@ var globalRegistry *vrfRegistry
 
 func init() {
 	globalRegistry = &vrfRegistry{
-		vrfs: make(map[string]*VRF),
+		vrfsName: make(map[string]*VRF),
+		vrfsID:   make(map[uint32]*VRF),
 	}
 }
 
 // vrfRegistry holds a reference to all active VRFs. Every VRF have to have a different name.
 type vrfRegistry struct {
-	vrfs map[string]*VRF
-	mu   sync.Mutex
+	vrfsName map[string]*VRF
+	mu       sync.Mutex
+	vrfsID   map[uint32]*VRF
 }
 
 // registerVRF adds the given VRF from the global registry.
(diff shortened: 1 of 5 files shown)


routingtable/vrf/vrf_registry.go
```diff
@@ -9,14 +9,16 @@ var globalRegistry *vrfRegistry
...
+		vrfsName: make(map[string]*VRF),
+		vrfsID:   make(map[uint32]*VRF),
 	}
 }
 
 // vrfRegistry holds a reference to all active VRFs. Every VRF have to have a different name.
 type vrfRegistry struct {
-	vrfs map[string]*VRF
-	mu   sync.Mutex
+	vrfsName map[string]*VRF
```
hikhvar: Why do we track the vrfs by booth name and id? Could we just use a single map[id]->VRF and check the name of the VRF with a given id.

cedi: but then we can't lookup a vrf by name.
I could do it the other way round to do the second lookup with `map[name]id` but if we wanna lookup by id in the future this could also lead to some problems.
Memory wise i don't think that's a big difference, since we are only storing `map[string] pointer to to VRF`

hikhvar: OK. Good arguments.

taktv6:
I don't see, why this ID is necessary in the VRF structure.
IMHO it should be contained in the FIB module. The FIB is a client of an AFI/SAFI (or table) of a VRF. How the FIB is programmed to a data plane is up to the FIB module and not of any concern for the VRF structure.

PR #199, opened 2019-03-09
Implement the FIB

cedi:
With this PR I'm adding the long awaited FIB functionality to serialize routes from the locRIB to a routing table in the kernel.
This implementation should be Platform independent. Currently there is one OS-Adapter for darwin that doesn't do anything besides being a stub for a future implementation. I don't have a mac to implement or test against this part. So I'm leaving this out for a future exercise ;-)
Please note, with this changes there is only the rib to kernel direction implemented. Currently no routes from the kernel are learned to the locRIB.
I wanted to split this functionality to keep the patch smaller and easier to review.
If there are requests for changes to the overall architecture in how i interact with the kernel it's easier to implement with a smaller patch and I should be more flexible to add the route learning feature later


cedi:
Please review (and merge) #198 first. This PR depends on changes from #198

routingtable/vrf/vrf.go
```diff
@@ -21,16 +21,22 @@ type addressFamily struct {
...
 	name     string
+	id       uint32
 	ribs     map[addressFamily]*locRIB.LocRIB
 	mu       sync.Mutex
 	ribNames map[string]*locRIB.LocRIB
 }
 
+// NewDefaultVRF creates a default VRF
+func NewDefaultVRF() (*VRF, error) {
+	return New("master", 254)
```
taktv6: This is linux specific and doesn't belong here.

PR #204, opened 2019-03-27
BGP and VRF metrics

czerwonk:
This PR does not include metrics for filtered and accepter routes because there is some basic stuff to extend before beeing able to implement this (new branch)


examples/bgp/main.go
```diff
@@ -23,7 +29,18 @@ func main() {
...
 	select {}
 }
+
+func startMetricsEndpoint(server server.BGPServer) {
+	prometheus.MustRegister(prom_bgp.NewCollector(server))
+	prometheus.MustRegister(prom_vrf.NewCollector())
+
+	http.Handle("/metrics", promhttp.Handler())
+	logrus.Info("Metrics are available :8080/metrics")
+	log.Fatal(http.ListenAndServe(":8080", nil))
```
hikhvar: Do we want to shut down the complete process if a auxiliary function like metrics fails? We should log the error. But the operator should decide how to act in case of an failed metrics endpoint.

czerwonk: It's the example. so I would not mind. In fact in the config based full featured routing daemon (which will come later) we should not terminate. In this case we should even implement a retry mechanism / auto heal. We can not afford to restart the daemon to get our metrics back. For the example file I will change it to log.Error without any further handling.

protocols/bgp/server/fsm_counter.go
```diff
@@ -0,0 +1,11 @@
+package server
+
+type fsmConters struct {
```
taktv6: Fix name.

czerwonk: Fixed

PR #218, opened 2019-07-28
Add caching for Path, BGPPath, IP and Prefix structures

taktv6:

PR #220, opened 2019-08-11
WIP: Implment mtr parser

hikhvar:
Currently a prototype for decoding MTR files according to https://tools.ietf.org/html/rfc6396.
The current scope is to implement the subtypes of TABLE_DUMP_V2. This enables us to parse RIS dumps provided by the RIPE. This is useful for lab setups.

Currently this branch is based uppon #218


hikhvar:
Rebased on master.

PR #223, opened 2019-08-19
Cleanup

corny:
this replaces #183


corny:
@czerwonk Ping

cmd/ris/risserver/server.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/bio-routing/bio-rd/routingtable"
 	"github.com/bio-routing/bio-rd/routingtable/filter"
 	"github.com/bio-routing/bio-rd/routingtable/locRIB"
+	"github.com/pkg/errors"
```
q3k: I would wait for Go2 errors, or replace `github.com/pkg/error` with something Go2-error-proposal compatible.

corny: Do you know a package that is closer to Go2-error-proposal?

q3k: https://godoc.org/golang.org/x/xerrors, which also claims to be partially merged into 1.13.

hikhvar (changes requested):
Please use fmt.Errorf with %w instead of github.com/pkg/errors

PR #224, opened 2019-08-22
Fix type assertions

taktv6:


protocols/bgp/packet/path_attributes.go
```diff
@@ -637,7 +637,7 @@ func (pa *PathAttribute) serializeCommunities(buf *bytes.Buffer) uint16 {
 		return 0
 	}
 
-	coms := pa.Value.(*[]uint32)
+	coms := pa.Value.(*types.Communities)
```
hikhvar: I think we should at regression tests to ensure the communities are serialized correctly.

taktv6: Absolutely. But not in this quick fix.

protocols/bgp/server/fsm_open_sent.go
```diff
@@ -137,6 +137,7 @@ func (s *openSentState) openMsgReceived(openMsg *packet.BGPOpen) (state, string)
 
 func (s *openSentState) handleOpenMessage(openMsg *packet.BGPOpen) (state, string) {
 	s.fsm.holdTime = time.Duration(math.Min(float64(s.fsm.peer.holdTime), float64(time.Duration(openMsg.HoldTime)*time.Second)))
+	fmt.Printf("s.fsm.holdTime: %d\n", s.fsm.holdTime)
```
czerwonk: in my opinion, this should be logging, but can be fixed later

taktv6: Fixed it.

PR #225, opened 2019-09-02
Implement OSPFv3 packet layer

mraerino:
Supports decoding and encoding of **OSPFv3 messages** in **IPv6 packets**.

**RFC:** [5340 - OSPF for IPv6](https://tools.ietf.org/html/rfc5340)

## Test suite

Testing is based on [packet captures from packetlife](https://packetlife.net/captures/protocol/ospf/) - in `fixtures`

### Decoding

Decodes all packets from the capture. Only fails on unexpected conditions (length or checksum mismatch).

### Encoding

Encodes go structs from snapshots in `fixtures/packets`. Compares the results to the packet capture.
To update the snapshots, run `go run ./protocols/ospf/packetv3/fixtures/packets/gen/`.


mraerino:
Superseded by https://github.com/bio-routing/bio-rd/pull/235

PR #228, opened 2019-09-28
BGP speed improvement

taktv6:


benchmarks/bgp/decode_real_full_feed/main.go
```diff
@@ -0,0 +1,144 @@
...
+	nRuns = flag.Int("runs", 1, "# runs")
+)
+
+type task struct {
+	num int
+	raw *bytes.Buffer
+	msg *packet.BGPMessage
+}
+
+func main() {
```
hikhvar: Let us use normal go benchmarks instead of own Programm. They support profiling too. https://medium.com/@felipedutratine/profile-your-benchmark-with-pprof-fb7070ee1a94

taktv6: My issue with that is, that they start measuring the whole function. I just want to measure certain actions on a populated data structure. If you know a way that integrates properly and allows this I'm open for change.

mraerino: we tried to refactor to benchmarks with a setup function but ran into issues with the buffers used. Since benchmarks are not part of the CI, we decided to no integrate it as a standard benchmark

net/ip.go
```diff
@@ -16,10 +16,24 @@ type IP struct {
...
+}
+
+// Copy creates a copy
+func (ip IP) Copy() *IP {
+	return &IP{
+		higher:   ip.higher,
+		lower:    ip.lower,
+		isLegacy: ip.isLegacy,
+	}
+}
```
mraerino: This is unused, right?

net/ip.go
```diff
@@ -124,7 +156,7 @@ func (ip *IP) Equal(other *IP) bool {
 }
 
 // Less compares ips for use in btree.Btree
-func (ip *IP) Less(other btree.Item) bool {
+func (ip IP) Less(other btree.Item) bool {
 	return ip.Compare(other.(*IP)) == -1
 }
```
mraerino: Is this still needed?

net/prefix.go
```diff
@@ -19,12 +19,25 @@ type Prefix struct {
...
+	return &p
+}
+
+// Copy creates a copy of the prefix
+func (p Prefix) Copy() *Prefix {
+	return &Prefix{
+		addr:   p.addr,
+		pfxlen: p.pfxlen,
+	}
+}
```
mraerino: Unused?

net/prefix.go
```diff
@@ -19,12 +19,25 @@ type Prefix struct {
...
+
 // Less compares prefixes for use in btree.Btree
-func (p *Prefix) Less(other btree.Item) bool {
+func (p Prefix) Less(other btree.Item) bool {
```
mraerino: Is this still needed?

net/prefix_cache.go
```diff
@@ -19,27 +17,29 @@ func init() {
...
 	}
 }
 
-func (pfxc *pfxCache) get(pfx *Prefix) *Prefix {
+func (pfxc *pfxCache) get(pfx Prefix) *Prefix {
 	pfxc.cacheMu.Lock()
 
-	item := pfxc.tree.Get(pfx)
-	if item != nil {
+	if p, exists := pfxc.cache[pfx]; exists {
```
mraerino: Should the IP in the prefix be deduped first, so we get more hits?

PR #229, opened 2019-09-28
Replace ip/prefix map caches with btree caches.

taktv6:


net/prefix_cache.go
```diff
@@ -1,9 +1,13 @@
...
-import "sync"
+import (
+	"sync"
+
+	"github.com/google/btree"
+)
 
 const (
-	prefixCacheInitialSize = 1000000
+	prefixCacheBTreeGrade = 3500
```
BarbarossaTM: why 3500?

net/prefix_cache.go
```diff
@@ -15,25 +19,26 @@ func init() {
...
 type pfxCache struct {
-	cache   map[Prefix]*Prefix
 	cacheMu sync.Mutex
+	tree    *btree.BTree
 }
 
 func newPfxCache() *pfxCache {
 	return &pfxCache{
-		cache: make(map[Prefix]*Prefix, prefixCacheInitialSize),
+		tree: btree.New(3500),
```
BarbarossaTM: why 3500?

net/ip_cache.go
```diff
@@ -1,9 +1,13 @@
...
-import "sync"
+import (
+	"sync"
+
+	"github.com/google/btree"
+)
 
 const (
-	ipCacheInitialSize = 1000000
+	ipCacheBTreeGrade = 3500
```
BarbarossaTM: why 3500?

net/prefix_cache.go
taktv6: Because it's very good!
Seriously: I've benchmarked this to be fastest on my machine :D

net/ip_cache.go
taktv6: Because it's very good!
Seriously: I've benchmarked this to be fastest on my machine :D

PR #231, opened 2019-10-11
Add prefix dedup function that also dedups IPs

taktv6:


net/prefix.go
```diff
@@ -22,6 +22,12 @@ func (p *Prefix) Dedup() *Prefix {
 	return pfxc.get(p)
 }
 
+// DedupWithIP gets a copy of Prefix from the cache and dedups the IP part
+func (p *Prefix) DedupWithIP() *Prefix {
+	p.addr = p.addr.Dedup()
```
hikhvar: Is it intended that this function alters the original Prefix?

taktv6: Yes. That's fine.

PR #251, opened 2020-04-11
Add support for setting TTL on BGP sessions

taktv6:


awlx:
LGTM

protocols/bgp/server/fsm_active.go
```diff
@@ -54,6 +54,12 @@ func (s *activeState) connectRetryTimerExpired() (state, string) {
 }
 
 func (s *activeState) connectionSuccess(con net.Conn) (state, string) {
+	if s.fsm.peer.ttl != 0 {
+		SetTCPConnTTLSockopt(con, s.fsm.peer.ttl)
+	} else if s.fsm.peer.isEBGP() {
+		SetTCPConnTTLSockopt(con, 1)
```
BarbarossaTM, line 60: This way we do not allow multihop eBGP, which breaks GRE-setups like FFRL.

PR #252, opened 2020-04-12
BGP: Add support for MD5 secret (RFC2385)

taktv6:


net/tcp/listen.go
```diff
@@ -0,0 +1,120 @@
...
+
+// SetTCPMD5 sets a TCP md5 secret for addr
+func (l *Listener) SetTCPMD5(addr net.IP, secret string) error {
+	isIPv4Listener := l.laddr.IP.To4() != nil
+	isIPv4Client := addr.To4() != nil
+
+	// Nothing to do if address families don't match
+	if isIPv4Client != isIPv4Listener {
+		return nil
+	}
```
BarbarossaTM, line 88: Please comment on why this is neccessary and 'peer_addr' instead of addr would be more clear.

net/tcp/md5sig.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type tcpMD5sig struct {
+	ssFamily  uint16
+	ss        [126]byte
+	flags     uint8
+	prefixLen uint8
+	keylen    uint16
+	ifIndex   uint32
+	key       [80]byte
```
BarbarossaTM, line 22: please use tcpMD5SIGMaxKeyLen instead of 80, because magic value

protocols/bgp/server/fsm.go
```diff
@@ -242,11 +243,38 @@ func (fsm *FSM) cease() {
 	fsm.eventCh <- Cease
 }
 
+func (fsm *FSM) sockSettings(c net.Conn) error {
+	ttl := fsm.peer.ttl
+	setNoRoute := false
+	if ttl == 0 {
+		if fsm.peer.isEBGP() {
```
BarbarossaTM, line 250: Requires user of eBGP multihop to set an TTL explicitly

taktv6: That's intended.

PR #253, opened 2020-04-14
Implement a GetRouters RPC to RIS

hikhvar:
This RPC returns all configured routers and their VRF IDs. This can be
used by RIS clients to discovery the routers. The router and the VRF ID
is a required parameter for all other RPCs.


net/api/net.pb.go
```diff
@@ -18,7 +26,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
+const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
```
hikhvar, line 29: I run the `regenerate_proto.sh` script. Maybe my protoc is too old :thinking:

PR #256, opened 2020-04-15
Use github actions

hikhvar:


corny (dismissed):

corny (approved):

corny (dismissed):

corny:
@taktv6 please merge

.github/linters/.golangci.yml
```diff
@@ -0,0 +1,3 @@
+run:
+  skip-dirs:
+    - net/tcp
```
hikhvar, line 3: Why do we exclude this?

corny (dismissed):

.github/linters/.golangci.yml
corny: I was not able to fix the errors. If you have some time, give it a try:

https://github.com/bio-routing/bio-rd/runs/1187348144?check_suite_focus=true

taktv6: So the new CI system would not build certain parts of the project. 
Sorry, wont merge.

corny: It's just the linter that has been added. It did not exist before.

taktv6:
Please make all actions work properly. Then I'm happy to merge.

awlx:
Well the actions work, but the linter is really unhappy. It basically complains about every file, maybe this one shouldn't be used.

awlx:
@taktv6 are we sure we want to test on MacOS or should that be excluded for now?

awlx:
The circleci Error should go away as soon as it's merged ... as it's gone then.

corny (dismissed):

corny:
@taktv6 please merge

.github/workflows/tests.yml
```diff
@@ -0,0 +1,32 @@
+on: [push, pull_request]
+name: tests
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-version: [1.16.x, 1.17.x]
```
taktv6, line 7: Please 1.17 only here. 1.16 would not build anymore.

taktv6 (approved):

PR #263, opened 2020-05-09
Add functions to AdjRIBIn/AdjRIBOut for RIS

sebageek:
These functions are required so AdjRIBIn and AdjRIBOut can be used the
same way as LocRIB. With this we can treat the RIB of a BMP server the
same way as the RIB of a BGP server.

---
To use the RIB of BMP and BGP in RIS I need to extend this interface. There is also the option of putting the Interface into the RIS or adding another interface to the routingtable package. In any case we would need to add `Get()`, `LPM()` and `GetLonger()` to be exposed.

I'm very open for input on this one.

### DIFF
--- a/routingtable/adjRIBIn/adj_rib_in.go
+++ b/routingtable/adjRIBIn/adj_rib_in.go
@@ -254,6 +254,11 @@ func (a *AdjRIBIn) Register(client routingtable.RouteTableClient) {
 	a.clientManager.RegisterWithOptions(client, routingtable.ClientOptions{BestOnly: true})
 }
 
+// RegisterWithOptions registers a client for updates
+func (a *AdjRIBIn) RegisterWithOptions(client routingtable.RouteTableClient, options routingtable.ClientOptions) {
+	a.clientManager.RegisterWithOptions(client, options)
+}
+
 // Unregister unregisters a client
 func (a *AdjRIBIn) Unregister(client routingtable.RouteTableClient) {
 	if !a.clientManager.Unregister(client) {
@@ -270,4 +275,19 @@ func (a *AdjRIBIn) Unregister(client routingtable.RouteTableClient) {
 // RefreshRoute is here to fultill an interface
 func (a *AdjRIBIn) RefreshRoute(*net.Prefix, []*route.Path) {
 
+}
+
+// LPM performs a longest prefix match on the routing table
+func (a *AdjRIBIn) LPM(pfx *net.Prefix) (res []*route.Route) {
+	return a.rt.LPM(pfx)
+}
+
+// Get gets a route
+func (a *AdjRIBIn) Get(pfx *net.Prefix) *route.Route {
+	return a.rt.Get(pfx)
+}
+
+// GetLonger gets all more specifics
+func (a *AdjRIBIn) GetLonger(pfx *net.Prefix) (res []*route.Route) {
+	return a.rt.GetLonger(pfx)
 }
--- a/routingtable/adjRIBOut/adj_rib_out.go
+++ b/routingtable/adjRIBOut/adj_rib_out.go
@@ -275,6 +275,11 @@ func (a *AdjRIBOut) Register(client routingtable.RouteTableClient) {
 	a.clientManager.RegisterWithOptions(client, routingtable.ClientOptions{BestOnly: true})
 }
 
+// RegisterWithOptions registers a client for updates
+func (a *AdjRIBOut) RegisterWithOptions(client routingtable.RouteTableClient, options routingtable.ClientOptions) {
+	a.clientManager.RegisterWithOptions(client, options)
+}
+
 // Unregister unregisters a client
 func (a *AdjRIBOut) Unregister(client routingtable.RouteTableClient) {
 	a.clientManager.Unregister(client)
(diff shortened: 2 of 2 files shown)


protocols/bgp/server/update_sender.go
```diff
@@ -383,3 +383,18 @@ func (a *UpdateSender) ReplacePath(*net.Prefix, *route.Path, *route.Path) {
 func (a *UpdateSender) RefreshRoute(*net.Prefix, []*route.Path) {
 
 }
+
+// LPM is here to fulfill an interface
+func (a *UpdateSender) LPM(pfx *net.Prefix) (res []*route.Route) {
+	return nil
```
BarbarossaTM, line 389: To be consistent with above there should be a warning like

`log.Warningf("BGP Update Sender: RouteCount not implemented")`

protocols/kernel/kernel.go
```diff
@@ -78,3 +78,18 @@ func (k *Kernel) ReplacePath(*net.Prefix, *route.Path, *route.Path) {
 func (k *Kernel) RefreshRoute(*net.Prefix, []*route.Path) {
 
 }
+
+// Get is here to fulfill an interface
+func (k *Kernel) Get(pfx *net.Prefix) *route.Route {
+	return nil
```
BarbarossaTM, line 84: This seems to be ok for now, but I guess the kernel RT needs some love.

protocols/bgp/server/update_sender.go
sebageek: Done

routingtable/client_interface.go
```diff
@@ -13,10 +13,14 @@ type RouteTableClient interface {
...
 	UpdateNewClient(RouteTableClient) error
 	Register(RouteTableClient)
+	RegisterWithOptions(RouteTableClient, ClientOptions)
 	Unregister(RouteTableClient)
 	RouteCount() int64
 	ClientCount() uint64
 	Dump() []*route.Route
 	ReplaceFilterChain(filter.Chain)
 	RefreshRoute(*net.Prefix, []*route.Path)
+	Get(pfx *net.Prefix) *route.Route
```
hikhvar, line 23: Why do we need to increase the surface of this interface? Is the interface used somewhere in the RIS? 
I prefer smaller interfaces. If we want to define a general interface to plugin the AdjRIBIn or AdjRIBOut into the RIS, we should define a smaller interface in the RIS package. In Go we should define the smallest interface as possible in the consumer package.

sebageek: I'm currently implementing a version of the ris that uses both BMP and BGP as a data source. For this I need to have an interface that covers LocRIB and AdjRIBIn, but this is not yet used anywhere in current master. Extending the RouteTableInterface was one possible option to solve my problem.

I talked to @taktv6 off-github about initially and he said he would be fine with this proposed solution, as he wants to refactor the RIB* implementations anyway. If you on the other hand do not feel comfortable with this I can change this PR to only add `RegisterWithOptions()`, `Get()`, `GetLonger()` and `LPM()` to AdjRIBIn (and for the sake of consistency to AdjRIBOut), leafe the RouteTableClient interface alone and create my own little interface with just these four functions I actualy need in the ris package.

hikhvar: Since it is no problem in golang to implement multiple interfaces I prefer a new smaller interface in the RIS package. 
> and for the sake of consistency to AdjRIBOut
It may be consistent but does it have any meaning? If AdjRIBOut is not a meaningful RIS table client, why support it?

sebageek: I don't need it ith AdjRIBOut, but it could be used there, as it is a RoutingTable member. For example one could ask the AdjRIBOut what routing would look like for the RIB advertised towards another peer. The "sake of consistency" was ment as "if I'm already adding those methods to AdjRINBIn (which I need for the RIS-case) I could also add them to AdjRIBOut (where I don't need them, but they could be useful to somebody else). Does that make sense? If you'd prefer to keep the PR minimal here as well I can also only explicitly add the function I require for my RIS-use-case.

hikhvar: Sorry, had the wrong struct in mind. The AdjRIBOut does make some sense. I thought of the update sender :facepalm:

sebageek: As title, commit message and code is a bit different I have opened a new PR as an alternative option. Does #268 look better? (If you prefer I can also push the changes into this PR)

hikhvar: I like the other MR! It keeps the RouteTableClient interface smaller. If there is a need to add more functions to the interface we can do it later.

sebageek: Cool :) Should I push the changes into this PR?

sebageek:
Moved changes from #268 to here and closed the alternative. As mentioned in the other PR, the interface for the could look something like this then (but I'd keep that in the RIS part of the code)
```golang
type RISRIB interface {
    RegisterWithOptions(routingtable.RouteTableClient, routingtable.ClientOptions)
    Unregister(routingtable.RouteTableClient)
    Dump() []*route.Route
    LPM(pfx *net.Prefix) (res []*route.Route)
    Get(pfx *net.Prefix) *route.Route
    GetLonger(pfx *net.Prefix) (res []*route.Route)
}
```

hikhvar (approved):

PR #269, opened 2020-05-12
Replace paths with same identifier with AddPath

sebageek:
RFC7911 sec 5 par 1 sais that the path identifier added by BGP Add Path
uniquely identifies a path for a given prefix. Therefore, when a new
path arrives we need to replace all paths with the same path identifier
for this prefix. Ignoring this leads to routes piling up in the
adjRIBIn.

### DIFF
--- a/protocols/bgp/server/fsm_address_family.go
+++ b/protocols/bgp/server/fsm_address_family.go
@@ -155,6 +155,7 @@ func (f *fsmAddressFamily) updates(u *packet.BGPUpdate) {
 	for r := u.NLRI; r != nil; r = r.Next {
 		path := f.newRoutePath()
 		f.processAttributes(u.PathAttributes, path)
+		path.BGPPath.PathIdentifier = u.NLRI.PathIdentifier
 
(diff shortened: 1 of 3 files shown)


sebageek:
There is still a problem with this PR. All tests work properly, but `adjRIBIn.AddPath()` does not see the PathIdentifiers, they are always 0, which renders AddPath broken again. I'll continue investigating.

sebageek:
I hope I've found the correct place to copy the PathIdentifier. Also it would be nice to write a breakfix test for this, but I have not found a good place in the codebase where other tests like this would be, e.g. testing a route advertisement with addpath enabled starting with the raw packet and checking the endresult in the adjInRIB.

sebageek:
Are there any integration tests in bio? Input "binary stream of bgp messages", check if the right routes are in the RIB?

hikhvar:
> Are there any integration tests in bio? Input "binary stream of bgp messages", check if the right routes are in the RIB?

I don't know the existens of such tests.

PR #270, opened 2020-05-16
Export neighbors in bmp server

sebageek:
I'm thinking about exposing the neighbors of a bmp session (and a bgp server later on). Neighbors needs to be public for that. Haven't tested it yet but wanted to ask you upstream people for an early opinion.

### DIFF
--- a/protocols/bgp/server/bmp_neighbor_manager.go
+++ b/protocols/bgp/server/bmp_neighbor_manager.go
@@ -6,17 +6,17 @@ import (
 )
 
 type neighborManager struct {
-	neighbors   []*neighbor
+	neighbors   []*Neighbor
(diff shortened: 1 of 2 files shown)


hikhvar:
I don't see a reason to not make neighbors public. However I see no value in the current draft. The neighbour structs is public but the members of that structs are neither public nor have getters. So only code from the bmp package can doe something useful with it. Hence we can just leave it private to the bmp package.

Did I miss something?

sebageek:
Yeah, you are right. My initial goal was to get the neighbors somehow exposed in the RIS, but I'd need to do this somehow for BMP and BGP. The already present GetSession API looks like it would be the way to go, exposing Neighbors directly does not.

Thanks for the comment!

PR #271, opened 2020-05-23
Implement ListSessions of BGP API

sebageek:
VRF filtering and some session stats are still missing here, but listing
sessions does work.

---

I'm not 100% on what I should fill in into the SessionStat, that's why I only filled them partly.


sebageek:
From a quick look it should also be easy to get RoutesImported and RoutesExported from the metrics service. I'll add this to this PR tomorrow.

sebageek:
I added some stuff. The session API now has a `vrf_name` field, the description field is filled and vrf_name filtering is implemented.

PR #278, opened 2020-07-15
RIS: Implement initial dump signaling in ObserveFIB

taktv6:


cmd/ris/risserver/server.go
```diff
@@ -251,6 +270,24 @@ func (s *Server) GetRouters(c context.Context, request *pb.GetRoutersRequest) (*
 	return resp, nil
 }
 
+type RequestWithVRF interface {
+	GetVrfId() uint64
+	GetVrf() string
+}
+
+func getVRF(req RequestWithVRF) (uint64, error) {
```
BarbarossaTM: This should be called getVRFID(), otherwise it's confusing why this returns an ID and not a string as getVrf about does.

routingtable/vrf/vrf.go
```diff
@@ -147,3 +150,28 @@ func RouteDistinguisherHumanReadable(rdi uint64) string {
...
+
+	netID, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return 0, errors.Wrap(err, "Unable to convert second part to int")
+	}
+
+	ret := uint64(0)
+	ret = uint64(asn)
+	ret = ret << 32
+	ret += uint64(netID)
```
BarbarossaTM, line 183: Does this need input validation as in a check that both strings are a 32bit value?

PR #284, opened 2020-09-16
Add convenience functions for AS Paths

taktv6:
Functionality is needed for Flowhouse where IP flows are annotated with src/dst ASN.


BarbarossaTM (approved):
LGTM

PR #286, opened 2020-10-07
move benchmarks into own repo

costela:
This should workd around golang/go#41708 when trying to push the module e.g. over [athens](https://github.com/gomods/athens)


taktv6 (approved):
LGTM

PR #287, opened 2020-10-08
Add IP masking function

taktv6:


net/ip.go
```diff
@@ -309,3 +311,32 @@ func (ip *IP) Next() *IP {
...
+	if ip.isLegacy {
+		ip.maskLastNBitsIPv4(n)
+		return ip
+	}
+
+	ip.maskLastNBitsIPv6(n)
+	return ip
+}
+
+func (ip *IP) maskLastNBitsIPv4(n uint8) {
```
BarbarossaTM, line 328: I'd add a check if n <= 32 here

net/ip.go
```diff
@@ -309,3 +311,32 @@ func (ip *IP) Next() *IP {
...
+	mask := uint64((math.MaxUint64 << (n)))
+	ip.lower = ip.lower & mask
+}
+
+func (ip *IP) maskLastNBitsIPv6(n uint8) {
```
BarbarossaTM, line 333: Same as above with <= 128 obviously

net/ip.go
taktv6: For what? Cutting off more than the address length returns 0.0.0.0 and is what I would expect. No?

BarbarossaTM (approved):
Yo.

PR #289, opened 2020-10-27
Implement handler for ethernet sockets

taktv6:


net/ethernet/bpf.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	p := (*reflect.SliceHeader)(unsafe.Pointer(&prog)).Data
+	switch wordWidth {
+	case 4:
+		buf.Write(bnet.BigEndianToLocal(convert.Uint32Byte(uint32(uintptr(p)))))
+	case 8:
+		buf.Write(bnet.BigEndianToLocal(convert.Uint64Byte(uint64(uintptr(p)))))
+	default:
+		panic("Unknown word width")
```
hikhvar, line 75: I think we should return an error here instead of triggering a panic.

net/ethernet/conn.go
```diff
@@ -0,0 +1,81 @@
...
+func (ea EthAddr) String() string {
+	return ea.address
+}
+
+// Network returns "ethernet"
+func (ea EthAddr) Network() string {
+	return "ethernet"
+}
+
+// Conn is a wrapper that implements net.Conn and sents packets to a particular destination
```
hikhvar, line 24: Let's make it more clear that a ethernet.Conn is write only

BarbarossaTM (approved):
LGTM

PR #290, opened 2020-11-01
Remove old IS-IS server code

taktv6:
Let's remove the legacy code before adding the new one in order to make the upcoming PRs more straight forward.


BarbarossaTM (approved):
Aye

PR #292, opened 2020-11-02
Routes deletion

mcindrich:
The bio-rd daemon is changed so that it is closed when other connected daemons are closed or shut down, using a new gRPC connection. Deleting routes when the daemon is closing is also added by removing paths from the RIB when shutting down.


cmd/bio-rd/main.go
```diff
@@ -17,22 +26,35 @@ import (
...
 	grpcKeepaliveMinTime = flag.Uint("grpc_keepalive_min_time", 1, "Minimum time (seconds) for a client to wait between GRPC keepalive pings")
 	metricsPort          = flag.Uint("metrics_port", 55667, "Metrics HTTP server port")
-	sigHUP               = make(chan os.Signal)
 	vrfReg               = vrf.NewVRFRegistry()
 	bgpSrv               bgpserver.BGPServer
-	runCfg               *config.Config
+	shutdownAddr         string
 )
 
+func init() {
```
taktv6, line 44: What's the point of this? If you need a certain config file name then just use the "-config.file" flag.

cmd/bio-rd/main.go
```diff
@@ -17,22 +26,35 @@ import (
...
+	cname := "bio-rd-" + hostname + ".yml"
+	log.Infof("Loading config " + cname)
+
+	configFilePath = flag.String("config.file", cname, "bio-rd config file")
+}
+
 func main() {
 	flag.Parse()
 
+	// config
```
taktv6, line 57: I don't think this comment is needed. The line below says exactly what it does.

cmd/bio-rd/main.go
```diff
@@ -86,19 +118,60 @@ func main() {
...
+	for sig := range sigCh {
+		log.Infof("Loading configuration")
 		newCfg, err := config.GetConfig(*configFilePath)
 		if err != nil {
 			log.Errorf("Failed to get config: %v", err)
 			continue
 		}
 
+		if sig != syscall.SIGHUP {
+			log.Infof("Received signal to STOP")
```
taktv6, line 135: This log message is fake news. A signal != syscall.SIGHUP was received. Not necessarily a SIGSTOP.

protocols/bgp/api/bgp.proto
```diff
@@ -26,8 +26,14 @@ message DumpRIBRequest {
...
+message SignalNotification {
+    int32 signal = 1;
+    string addr = 2;
+}
+
 service BgpService {
     rpc ListSessions(ListSessionsRequest) returns (ListSessionsResponse) {}
     rpc DumpRIBIn(DumpRIBRequest) returns (stream bio.route.Route) {}
     rpc DumpRIBOut(DumpRIBRequest) returns (stream bio.route.Route) {}
+    rpc SignalChat(stream SignalNotification) returns (stream SignalNotification) {}
```
taktv6, line 38: Can you please explain why a client would need to know what signal the bio-rd has received?

taktv6 (commented):
This looks very weird to me.
Why should bio-rd create a GRPC connection it's BGP peers?
And why would it stream notifications as a streaming RPC?
I can't follow. Would you mind to please explain what you're going to achieve?
Thanks!

PR #297, opened 2021-01-20
Tear down ObserveRIB streaming RPC when RIB is destroyed

taktv6:
Background: After a router went down all routes have been withdrawn as expected to all clients currently running the ObserveRIB gRPC. However, the BMP server creates fresh VRFs and RIBs when a BMP session comes back up.
Thus the running ObserveRIB calls were using stale RIB references and in the end not sending any new routes received via BMP to the running ObserveRIB clients. It is better to just detect this situation and end the RPC so clients can reconnect and get information from the current RIB.

### DIFF
--- a/routingtable/client_interface.go
+++ b/routingtable/client_interface.go
@@ -13,7 +13,8 @@ type RouteTableClient interface {
 	RemovePath(*net.Prefix, *route.Path) bool
 	ReplacePath(*net.Prefix, *route.Path, *route.Path)
 	RefreshRoute(*net.Prefix, []*route.Path)
-	Destroy()
+	// A call to Dispose() signals that no more updates are to be expected from the RIB the client is registered to.
+	Dispose()
 }
 
 type AdjRIB interface {
@@ -38,5 +39,6 @@ type AdjRIBOut interface {
 	AddPathInitialDump(pfx *net.Prefix, path *route.Path) error
 	ReplacePath(*net.Prefix, *route.Path, *route.Path)
 	RefreshRoute(*net.Prefix, []*route.Path)
-	Destroy()
+	// A call to Dispose() signals that no more updates are to be expected from the RIB the client is registered to.
+	Dispose()
 }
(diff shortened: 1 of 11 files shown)


routingtable/vrf/vrf_registry.go
```diff
@@ -59,12 +59,15 @@ func (r *VRFRegistry) UnregisterVRF(v *VRF) {
 	delete(r.vrfs, v.routeDistinguisher)
 }
 
-// UnregisterAll unregisters all VRFs
-func (r *VRFRegistry) UnregisterAll() {
+// DisposeAll dosposes all VRFs
+func (r *VRFRegistry) DisposeAll() {
```
hikhvar, line 63: Maybe call this `DestroyAll` or rename the method of the interface to `Dispose`?

routingtable/client_interface.go
```diff
@@ -37,4 +38,5 @@ type AdjRIBOut interface {
 	AddPathInitialDump(pfx *net.Prefix, path *route.Path) error
 	ReplacePath(*net.Prefix, *route.Path, *route.Path)
 	RefreshRoute(*net.Prefix, []*route.Path)
+	Destroy()
```
hikhvar, line 41: What semantics are bound to this request? What should an implementor of this interface guarantee after the `Destroy` method is called?

routingtable/client_interface.go
```diff
@@ -13,6 +13,7 @@ type RouteTableClient interface {
 	RemovePath(*net.Prefix, *route.Path) bool
 	ReplacePath(*net.Prefix, *route.Path, *route.Path)
 	RefreshRoute(*net.Prefix, []*route.Path)
+	Destroy()
```
hikhvar, line 16: Same here. What are the semantics? Maybe move this function to an own interface and use interface compositioning?

hikhvar (changes requested):

routingtable/vrf/vrf_registry.go
taktv6: Done.

routingtable/client_interface.go
taktv6: Added comment

routingtable/client_interface.go
taktv6: Added comment

hikhvar (approved):

PR #299, opened 2021-01-22
BMP: Enable TCP keepalive

taktv6:


protocols/bgp/server/bmp_server.go
```diff
@@ -90,6 +92,14 @@ func (b *BMPServer) AddRouter(addr net.IP, port uint16) {
 				continue
 			}
 
+			if b.keepalivePeriod != 0 {
```
BarbarossaTM, line 95: This seems to only be not true if `keepalivePeriod` explicitly is set to 0 (as it's defaulted to 1 in the vars). Is this intended?

taktv6: Yes that is intended because having stale TCP sessions to BMP instances causes stale routing information.
When a routers loses power or gets isolated the BMP session would stay "up" for quite some time and network does not converge. Tested for you ;-)

BarbarossaTM (approved):
LGTM

PR #300, opened 2021-02-06
Add test for IPv4FromBytes

jayay:
This tests the code **the way it behaves right now**, without any modification. I assume it is correct, but I'm asking you to take good attention during the review. This one seems to be straight forward, but I will submit more PRs I'll need a little bit of help with 😬

### DIFF
--- a/net/ip_test.go
+++ b/net/ip_test.go
@@ -674,3 +674,24 @@ func TestMaskLastNBits(t *testing.T) {
 		assert.Equal(t, test.expected, res, test.name)
 	}
 }
+
+func TestIPv4FromBytes(t *testing.T) {
+	resultZero := IPv4FromBytes([]byte{})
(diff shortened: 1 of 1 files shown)


net/ip_test.go
```diff
@@ -674,3 +674,24 @@ func TestMaskLastNBits(t *testing.T) {
 		assert.Equal(t, test.expected, res, test.name)
 	}
 }
+
+func TestIPv4FromBytes(t *testing.T) {
+	resultZero := IPv4FromBytes([]byte{})
```
hikhvar, line 679: a more idomatic way to write this test would have been to use a table test like in `TestMaskLastNBits`. Since the testcases are very stable I will merge this anyways. No need to overengineer the tests.

hikhvar (approved):

PR #301, opened 2021-02-06
Test PathAttribute.Copy()

jayay:
This tests the code **the way it behaves right now**, without any modification. I assume it is correct, but I'm asking you to take good attention during the review.

This tests checks for the outcome to be a shallow copy.


protocols/bgp/packet/path_attributes_test.go
```diff
@@ -2383,3 +2383,37 @@ func TestDecode4ByteASN(t *testing.T) {
 		assert.Equal(t, test.expected, res, test.name)
 	}
 }
+
+func TestPathAttributeCopy(t *testing.T) {
+	attr := &PathAttribute{
```
hikhvar, line 2388: Can you make up a PathAttribute with every Field not set to a default value? This ensures every field is copied and not set to the default value.

hikhvar (dismissed):

protocols/bgp/packet/path_attributes_test.go
jayay: I've added every field that exists for now. However, the test will not fail if someone adds a new field to the struct and forgets to copy its value in the `Copy()` function. Do you think I should test that scenario using reflection or does that sound like a bad, over-engineered idea?

taktv6 (approved):
LGTM

PR #303, opened 2021-02-07
Add test for as_path.go + repair String()

jayay:
Adds test for `protocols/bgp/types/as_path.go`. The `String()` function of `ASPath` was re-implemented because its output was broken: sometimes spaces were missing, or in case it only contained one segment of type `ASSet`, it used to generate a leading space.

The rewrite of the `String` func was reviewed by @taktv6.

### DIFF
--- a/protocols/bgp/types/as_path.go
+++ b/protocols/bgp/types/as_path.go
@@ -1,7 +1,8 @@
 package types
 
 import (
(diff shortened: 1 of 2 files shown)


taktv6 (approved):
LGTM. Thanks for your contribution!

PR #308, opened 2021-11-28
add passive mode for BMP

tobikris:
Hey, I had a usecase for a passive BMP connection (e.g. the router should initiate the TCP connection).

I tried to add it to the best of my understanding of the inner workings of bio-rd.
I am not sure how to add a unit test for it though.

Side-note: one of the existing test cases is failing even on master (TestFSMAFIInitDispose).

### DIFF
--- a/protocols/bgp/server/bmp_server.go
+++ b/protocols/bgp/server/bmp_server.go
@@ -54,6 +54,7 @@ func NewServer(keepalivePeriod time.Duration) *BMPServer {
 }
 
 // Listen starts a listener for routers to start the BMP connection
+// the listener needs to be closed by calling Close() on the BMPServer
 func (b *BMPServer) Listen(addr string) error {
 	tcp, err := net.ResolveTCPAddr("tcp", addr)
 	if err != nil {
@@ -72,7 +73,6 @@ func (b *BMPServer) Listen(addr string) error {
 		return err
 	}
 	b.listener = l
-	defer l.Close()
 
 	for {
 		c, err := l.Accept()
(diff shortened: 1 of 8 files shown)


protocols/bgp/server/bmp_server.go
```diff
@@ -52,15 +53,116 @@ func NewServer(keepalivePeriod time.Duration) *BMPServer {
...
+		}
+		go b.handleConnection(c)
+	}
+}
+
+func (b *BMPServer) handleConnection(c net.Conn) {
+	remoteIP := c.RemoteAddr().(*net.TCPAddr).IP
+
+	var r *Router
+	for _, router := range b.getRouters() {
```
taktv6, line 94: Let's use the GetRouter() method here. Then just check for passive flag.

protocols/bgp/server/bmp_server.go
```diff
@@ -52,15 +53,116 @@ func NewServer(keepalivePeriod time.Duration) *BMPServer {
...
+			"router":         r.Name(),
+		}).Error("router already connected, dropping connection")
+		c.Close()
+		return
+	}
+
+	b.establishConnection(c, r, true)
+}
+
+func (b *BMPServer) establishConnection(c net.Conn, r *Router, passive bool) error {
```
taktv6, line 121: Not happy with this naming. Let's rename establishConnection to handleConnection. The current handleConnection should be validateConnection() bool. Move the call of establishConnection into the Listen() method and call it only if validateConnection() returned true.

protocols/bgp/server/bmp_server.go
tobikris: Thanks, changed it.

protocols/bgp/server/bmp_server.go
tobikris: I changed the names and moved the call.

taktv6 (approved):

PR #309, opened 2021-11-28
add rib filtering

tobikris:
I would like to be able to filter the RIB dump for the originating AS and prefix lengths.


route/route_test.go
```diff
@@ -666,3 +668,57 @@ func TestRouteEqual(t *testing.T) {
 		assert.Equal(t, tc.equal, res)
 	}
 }
+
+func TestRouteIsBGPOriginatedBy(t *testing.T) {
+	tests := []struct {
+		r      *Route
```
taktv6, line 674: Let's add a name/description to each test case.

route/route_test.go
```diff
@@ -666,3 +668,57 @@ func TestRouteEqual(t *testing.T) {
...
+				},
+			},
+			origBy: 65000,
+			isOrig: true,
+		},
+	}
+
+	for _, tc := range tests {
+		res := tc.r.IsBGPOriginatedBy(tc.origBy)
+		assert.Equal(t, tc.isOrig, res)
```
taktv6, line 722: Let's print the test cases name here too.

tobikris:
I added test case names and another test for an empty AS path.

taktv6 (approved):

PR #314, opened 2021-12-02
Drop linter

taktv6:


awlx:
Not sure if that’s enough but if not I have a look tomorrow.

awlx (approved):

PR #319, opened 2022-01-20
fix BGPPath Origin cases for string and print

mordner:
IGP and Incomplete cases are mixed up for string and print.


taktv6 (dismissed):
LGTM. Thanks!

taktv6:
I think you need to adjust this test: https://github.com/bio-routing/bio-rd/blob/master/route/bgp_path_test.go#L546

taktv6 (approved):
LGTM

PR #320, opened 2022-01-20
bmp: set local ASN to number from capability in case of ASN4

mordner:
Otherwise the localASN stays at 23456 and BGPPathA.EBGP is incorrectly reported as external for all routes.


taktv6 (approved):
LGTM

PR #328, opened 2022-07-10
ISIS: Allow removing interface

taktv6:


protocols/isis/server/server.go
```diff
@@ -81,3 +83,13 @@ func netsCompatible(nets []*types.NET) bool {
 
 	return true
 }
+
+// GetInterfaces gets all configured interfaces
+func (s *Server) GetInterfaces() []string {
```
BarbarossaTM, line 88: I'm wondering if this should be called `GetInterfaceNames` as that's what we return back. As all other public methods only deal with the names it's consistent though, so no strong feelings.

cmd/bio-rd/is-is.go
```diff
@@ -34,7 +34,13 @@ func configureProtocolsISIS(isis *config.ISIS) error {
 		}
 	}
 
+	configuredInterfaces := isisSrv.GetInterfaces()
 	for _, ifa := range isis.Interfaces {
+		if strSliceContains(configuredInterfaces, ifa.Name) {
+			log.Debugf("ISIS: Interface %q is already configured", ifa.Name)
```
BarbarossaTM, line 40: Is this method to be intended to be idempotent?

protocols/isis/server/server.go
taktv6: Fixed.

cmd/bio-rd/is-is.go
taktv6: That is the idea.

BarbarossaTM (approved):
LGTM

PR #329, opened 2022-07-17
Implement LSP RX handling

taktv6:
Implementing ISIS LSP RX functionality.

### DIFF
--- a/cmd/bio-rd/main.go
+++ b/cmd/bio-rd/main.go
@@ -37,6 +37,8 @@ var (
 func main() {
 	flag.Parse()
 
+	log.SetLevel(log.DebugLevel)
+
 	startCfg, err := config.GetConfig(*configFilePath)
 	if err != nil {
 		log.Errorf("unable to get config: %v", err)
--- a/protocols/isis/server/lsdb.go
+++ b/protocols/isis/server/lsdb.go
@@ -127,7 +127,7 @@ func (l *lsdb) sendLSPDUs() {
 	}
 }
 
-func (l *lsdb) processCSNP(csnp *packet.CSNP, from *netIfa) {
+func (l *lsdb) processCSNP(from *netIfa, csnp *packet.CSNP) {
 	l.lspsMu.Lock()
 	defer l.lspsMu.Unlock()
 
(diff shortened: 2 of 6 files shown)


cmd/bio-rd/main.go
```diff
@@ -37,6 +37,8 @@ var (
 func main() {
 	flag.Parse()
 
+	log.SetLevel(log.DebugLevel)
```
BarbarossaTM, line 40: Did you want to commit this? :)

protocols/isis/server/lsdb.go
```diff
@@ -312,3 +312,39 @@ func (l *lsdb) sendCSNPs(ifa *netIfa) {
...
+	defer l.lspsMu.Unlock()
+
+	existingLSDBEntry, exists := l.lsps[lspdu.LSPID]
+	if !exists || lspdu.SequenceNumber > existingLSDBEntry.lspdu.SequenceNumber {
+		log.Debugf("ISIS: Received newer LSPDU %v sequence number %d", lspdu.LSPID, lspdu.SequenceNumber)
+		l.processNewerLSPDU(ifa, lspdu)
+		return
+	}
+
+	if lspdu.SequenceNumber == existingLSDBEntry.lspdu.SequenceNumber {
```
BarbarossaTM, line 328: This could use an `exists` guard to be safe, and potentially an "I cannot let you do this Dave" case for an invalid LSPDU

protocols/isis/server/net_ifa.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	allISNetworkEntitiesAddr = ethernet.MACAddr{0x00, 0x2B, 0x00, 0x00, 0x05}
+	allISNetworkEntitiesAddr = ethernet.MACAddr{0x09, 0x00, 0x2B, 0x00, 0x00, 0x05}
```
BarbarossaTM, line 16: Out of curiosity: Did we only have the last 5 blocks here for a reason before? :-)

BarbarossaTM (changes requested):
I think this would be better with one more safe guard :)

cmd/bio-rd/main.go
taktv6: Sure, as long as this is under development and far away from a beta release.

protocols/isis/server/lsdb.go
taktv6: This is safe guarded. If !exists it will exit above.

BarbarossaTM (approved):
LGTM

protocols/isis/server/net_ifa.go
taktv6: As MAC addresses are usually 6 bytes I guess I missed a byte here when I initially wrote this.

PR #335, opened 2022-07-26
Allow BMP server to accept connections from any client

taktv6:


protocols/bgp/server/bmp_server.go
```diff
@@ -91,9 +98,16 @@ func (b *BMPServer) Listen(addr string) error {
 			}).Infof("Unable to accept on %s", tcp.String())
 			return err
 		}
+
+		if b.acceptAny {
+			tcpRemoteAddr := c.RemoteAddr().(*net.TCPAddr)
+			b.AddRouter(tcpRemoteAddr.IP, tcp.AddrPort().Port(), true)
```
BarbarossaTM, line 104: I believe we should `continue` the for loop here or guard the following `if` with ans `else`. Although we are accepting the `validateConnection` would log that we didn't :-)

BarbarossaTM (changes requested):

PR #344, opened 2022-07-31
Groundwork to store a reason why is route is hidden

BarbarossaTM:
(BGP) Routes can be ineligible for many reasons. This allows to store why
  we considered a given path as ineligible and are ignoring it.


route/path.go
```diff
@@ -2,15 +2,27 @@ package route
...
 import (
 	"fmt"
+	"strings"
 
 	bnet "github.com/bio-routing/bio-rd/net"
 	"github.com/bio-routing/bio-rd/route/api"
 )
 
+const (
+	HiddenReasonNone               = 0
```
taktv6, line 12: iota pls

taktv6 (approved):
LGTM

PR #347, opened 2022-08-05
Store the unix epoch when a route is learned (for BGP and BMP for now).

BarbarossaTM:
This allows storing the unix timestamp (epoch) of when a route is learned.
When a route is learned via BMP the timestamp provided by BMP is used (and
the microseconds ignored), and when the route is learned via BGP the local
time of that moment is used.


taktv6 (approved):
LGTM

PR #363, opened 2022-08-19
Abstract logger

### LINKED ISSUE #362: Allow a user to specify its own logger
When using bio-rd as a library, this is not possible to override the global logrus logger (well it is, as long as you only have one dependency using logrus, but that's not super clean). Instead, each component should use its own logger. `bgp/server.Router` already has its own logger, but it is not user-settable.

I would propose to add a `Logger` field to each structure needed it. It will be set to `log.New()` by default or inherited if it is sensible. As long as the "New" methods don't try to log, this should work good enough. This does not require modifying any signature.

I can do the work if it is deemed acceptable. More complex alternatives would be to define a logging interface to avoid depending on logrus, but I don't need that myself and this is far more work.

vincentbernat:
Replace direct to Logrus with a proxy exposing an abstract interface
very close to Logrus.

The changes include:

- add the `util/log` module
- replace the import of logrus with the import of this module
- call `SetLogger` at the main entry point (not strictly necessary)
- use `logrus` explicitely at some point

Also, the BMP server is losing its internal logger. It was used for
tests, but these tests are commented. The test has been updated to use
`SetLogger` instead (but it has not been really tested, ideally, it
should be a text fixture undoing the change once the test has run).

Fix #362.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,12 +20,13 @@ coverage.txt
 
 # 'go build' binary
 /cmd/bio-rd/bio-rd
-examples/bgp
-examples/bmp
-examples/fib/fib
-examples/device/device
-cmd/ris/ris
-benchmarks/bgp/learning/learning
+/cmd/bio-rdc/bio-rdc
+/cmd/ris-mirror/ris-mirror
+/cmd/ris/ris
+/cmd/riscli/riscli
+/examples/bgp/bgp
+/examples/bmp/bmp
+/examples/kernel
 
 # bazel directories
 /bazel-*
--- a/README.md
+++ b/README.md
@@ -23,12 +23,6 @@ cd examples/bgp/ && go build
 cd examples/bmp/ && go build
 ```
 
-#### Device
-
-```bash
-cd examples/device && go build
-```
-
 ### Run Tests
 
 ```bash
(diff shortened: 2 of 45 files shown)


hikhvar:
I like this. I think the cost of having a global singleton, are outweighed by the benefits of having a consistent user definable logging. 

I would maybe question if we need so much different log levels. I mainly have issues with the `Warning` log level. Personally, I think `Error`, `Info` and `Debug` should be enough. Moreover, `log.Fatal` should not be used anywhere within the library part of bio-routing. This should only be used within the toplevel `main` functions. Here, we already have a concrete logrus logger at hand. Therefore, I would argue to remove the `Fatal` log level as well.

vincentbernat:
Yes, we can remove the `Fatal()`. It seems used at one place in the library and I think this is an error.

vincentbernat:
I have removed the extra levels and avoid using `Fatal()` at all. In `main.go`, they are replaced by `Error()` + `os.Exit(1)`. The `Warning()` calls have been changed to either `Error()` or `Info()`.

taktv6 (approved):

taktv6:
Thanks for contributing. Really appreciated!

PR #364, opened 2022-08-21
BMP: Add a method to get listener local address

vincentbernat:
This is useful when listening on port 0 for testing purpose. We can
get the free port the kernel has allocated to us.


taktv6:
One can listen on port 0 and get a free port assigned? I'm 20 years into this Linux thingy but I didn't know that trick. Nice :)

taktv6 (approved):

hikhvar:
TIL also that. Nice!

PR #366, opened 2022-08-23
Rework the IP datatype to be a value by default and change prefix to contain an IP and not a pointer

taktv6:
We used to try do deduplicate IP addresses and prefixes as good as possible while doing so in a very controlled way.
This change rolls back the implicit deduplications happening in Prefix objects. I think in generals that we should handle IP addresses as plain objects whenever possible and only use pointers when we actually need the information to be deduplicated for memory efficiency reason. And that is actually only necessary in the routing tables. Path attributes are themselves deduplicated and to keep it simple we should not deduplicate the pieces that make up the path attributes.

### DIFF
--- a/net/ip.go
+++ b/net/ip.go
@@ -27,8 +27,8 @@ func (ip IP) Ptr() *IP {
 }
 
 // IPFromProtoIP creates an IP address from a proto IP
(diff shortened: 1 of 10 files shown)


BarbarossaTM (approved):
LGTM

PR #379, opened 2022-09-07
BMP: Allow ignoring BGP sessions based on peer ASN

taktv6:
Not sure yet if I really want this.


BarbarossaTM (commented):
LGTM, test would be nice

BarbarossaTM (approved):
LGTM

PR #380, opened 2022-09-12
BMP: Implement serialization of BMP messages

taktv6:


protocols/bmp/packet/initiation_message_test.go
```diff
@@ -18,6 +18,86 @@ func TestInitiationMessageMsgType(t *testing.T) {
...
+					},
+					{
+						InformationType: 2,
+						Information:     []byte("edge01.xyz01"),
+					},
+				},
+			},
+			expected: []byte{
+				0x3,                 // version
+				0x0, 0x0, 0x0, 0xe7, // message length (FIXME!)
```
BarbarossaTM, line 52: What's the FIXME?

BarbarossaTM (commented):
Can we somehow also validate the length? It should be at the same place in all bytes, so we should be able to compare `len(bytes) - 6` with int value of `bytes[1:5]` or something?

BarbarossaTM (approved):
LGTM

PR #382, opened 2022-09-13
Reflect Golang net.IP changes

BarbarossaTM:
It looks like Golang changed the behavior of the net.IP type with the latest
  redesign, so that the address family cannot be determined anymore by looking
  at the length of the underlying byte slice.  Adding two more tests to notice
  any regressions early in the future.


joao-reis (approved):
1.18 release note:
https://tip.golang.org/doc/go1.18#netip

PR #385, opened 2022-09-14
Rename BMP Server to BMP receiver in preparation for BMP Monitoring Station support

BarbarossaTM:
This is a preparation to implement a BMP sender :)


taktv6 (approved):
Trivial.

PR #394, opened 2022-11-25
fix: do not reconnect bgp from passive side

tobikris:
After the reconnectInterval even the passive connection side tried reconnecting.

### DIFF
--- a/protocols/bgp/server/fsm.go
+++ b/protocols/bgp/server/fsm.go
@@ -95,7 +95,7 @@ func NewPassiveFSM(peer *peer, con *net.TCPConn) *FSM {
 	return fsm
 }
 
-// NewActiveFSM initiates a new passive FSM
+// NewActiveFSM initiates a new active FSM
 func NewActiveFSM(peer *peer) *FSM {
 	fsm := newFSM(peer)
 	fsm.active = true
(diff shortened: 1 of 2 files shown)


protocols/bgp/server/fsm_idle.go
```diff
@@ -16,7 +16,7 @@ func newIdleState(fsm *FSM) *idleState {
 }
 
 func (s idleState) run() (state, string) {
-	if s.fsm.peer.reconnectInterval != 0 {
+	if !s.fsm.peer.passive {
```
BarbarossaTM, line 19: Hm, I'm wondering if we need to keep the check for `!= 0` here, too. @taktv6 any thoughts?

tobikris: I think this is a good suggestion, thanks.

protocols/bgp/server/fsm.go
```diff
@@ -95,7 +95,7 @@ func NewPassiveFSM(peer *peer, con *net.TCPConn) *FSM {
 	return fsm
 }
 
-// NewActiveFSM initiates a new passive FSM
```
taktv6, line 98: It looks like it was a typo but I think it was not. The BGP active state means the things is actively waiting for incoming connections and thus is passive.

From RFC4271:
 Active State:

      In this state, BGP FSM is trying to acquire a peer by listening
      for, and accepting, a TCP connection.

taktv6 (approved):

PR #396, opened 2022-11-26
Move contributing ASNs to VRF & add check for all Cluster IDs in the same VRF

BarbarossaTM:


protocols/bgp/server/fsm_address_family.go
```diff
@@ -138,7 +140,7 @@ func (f *fsmAddressFamily) getSessionAttrs() routingtable.SessionAttrs {
 }
 
 func (f *fsmAddressFamily) bmpInit() {
-	f.adjRIBIn = f.fsm.peer.adjRIBInFactory.New(filter.NewAcceptAllFilterChain(), &routingtable.ContributingASNs{}, f.getSessionAttrs())
+	f.adjRIBIn = f.fsm.peer.adjRIBInFactory.New(filter.NewAcceptAllFilterChain(), f.fsm.peer.vrf, f.getSessionAttrs())
```
BarbarossaTM, line 143: Should we just pass a freshly create VRF here to make 100% sure to adhere to what the router sent?

protocols/bgp/server/fsm_address_family.go
```diff
@@ -148,7 +150,7 @@ func (f *fsmAddressFamily) bmpInit() {
 }
 
 func (f *fsmAddressFamily) bmpDispose() {
-	f.rib.GetContributingASNs().Remove(f.fsm.peer.localASN)
+	f.vrf.RemoveContributingASN(f.fsm.peer.localASN)
```
BarbarossaTM, line 153: If we use a new VRF above, I guess this can go away?

protocols/bgp/server/fsm_address_family.go
```diff
@@ -44,6 +46,7 @@ func newFSMAddressFamily(afi uint16, safi uint8, family *peerAddressFamily, fsm
 		afi:               afi,
 		safi:              safi,
 		fsm:               fsm,
+		vrf:               fsm.peer.vrf,
```
taktv6, line 49: Why don't we access the vrf using fsm.peer.vrf but make a copy of the pointer here?

protocols/bgp/server/fsm_address_family.go
taktv6: I don't think so. The VRF should point to a VRF that was created for the specific RD that was in a peer up notification.

taktv6 (approved):

PR #403, opened 2022-12-08
Allow setting default Local Preference for eBGP sessions and default it to 100.

### LINKED ISSUE #400: BGP: Local Pref 100 not set as default for received eBGP routes
**Describe the bug**
When a route from an eBGP neighbor is received and no policies are configured for this neighbor, the Local-Pref will be set to 0. Without any policies configured the Local-Pref should instead be set to 100, which is the correct default value.

**Steps to Reproduce**
Build `examples/bgp` and peer with another BGP speaker which has a different AS. Announce a route from the other BGP speaker towards bio-rd.

**Expected behavior**
I expect bio-rd to set a default Local-Pref of 100 for all received routes which do not have Local-Pref set, before any policies are evaluated.

**Configuration used**
Peer config for `examples/bgp`:
```
func addPeersIPv4(b server.BGPServer, v *vrf.VRF) {
	b.AddPeer(server.PeerConfig{
		AdminEnabled:      true,
		LocalAS:           65200,
		PeerAS:            65300,
		PeerAddress:       bnet.IPv4FromOctets(127, 0, 0, 1).Ptr(),
		LocalAddress:      bnet.IPv4FromOctets(127, 0, 0, 1).Ptr(),
		ReconnectInterval: time.Second * 15,
		HoldTime:          time.Second * 90,
		KeepAlive:         time.Second * 30,
		Passive:           true,
		//RouterID:          b.RouterID(),
		RouterID: uint32(1),
		IPv4: &server.AddressFamilyConfig{
			//ImportFilterChain: filter.NewAcceptAllFilterChain(),
			ExportFilterChain: filter.NewAcceptAllFilterChain(),
			AddPathSend: routingtable.ClientOptions{
				MaxPaths: 10,
			},
		},
		RouteServerClient: true,
		VRF:               v,
	})
}
```

Other BGP speaker was [gobgp](https://github.com/osrg/gobgp). Config:
```
[global.config]
  as = 65300
  router-id = "192.168.255.1"
  port = 1790

[[neighbors]]
  [neighbors.config]
    neighbor-address = "127.0.0.1"
    peer-as = 65200
  [neighbors.transport.config]
    local-address = "127.0.0.1"
    remote-port = 179
```
Inject route:
```
$ gobgp global rib add 192.168.1.1/32 -a ipv4
```
Check route in bio-rd:
```
$ go run ./cmd/bio-rdc/main.go -bio-rd localhost:1337 -cmd "show routes receive-protocol bgp 127.0.0.1"
192.168.1.1/32:
All Paths:
        Protocol: BGP
        Hidden: no
                Local Pref: 0
                Origin: Incomplete
                AS Path: 65300
                BGP type: external
                NEXT HOP: 127.0.0.1
                MED: 0
                Path ID: 0
                Source: 127.0.0.1
```
Local-Pref should be 100 not 0.

BarbarossaTM:
RFC4277 Section 8. states that 100 is a - if not the - common value for Local Preference on eBGP sessions, where it would be 0 if none is set.  To increase interoperability with other BGP speakers we now allow setting a default value for Local Preference for the BGP server and default to 100 if none was set.

Closes #400

### DIFF
--- a/protocols/bgp/server/server.go
+++ b/protocols/bgp/server/server.go
@@ -17,14 +17,24 @@ import (
 )
 
 const (
-	BGPVersion = 4
+	BGPVersion            = 4
+	DefaultLocaPreference = 100
 )
 
+type BGPServerConfig struct {
+	// Mandatory attributes
+	RouterID         uint32
+	ListenAddrsByVRF map[string][]string
+	DefaultVRF       *vrf.VRF
+
+	// Optional attributes
+	DefaultLocalPreference *uint32
+}
+
 type bgpServer struct {
+	config          BGPServerConfig
 	listenerManager tcp.ListenerManagerI
-	defaultVRF      *vrf.VRF
 	peers           *peerManager
-	routerID        uint32
 	metrics         *metricsService
 }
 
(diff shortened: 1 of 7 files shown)


protocols/bgp/server/server.go
```diff
@@ -45,10 +55,15 @@ type BGPServer interface {
 
 // NewBGPServer creates a new instance of bgpServer
 func NewBGPServer(routerID uint32, defaultVRF *vrf.VRF, listenAddrsByVRF map[string][]string) BGPServer {
-	return newBGPServer(routerID, defaultVRF, listenAddrsByVRF)
+	return newBGPServer(routerID, defaultVRF, listenAddrsByVRF, DefaultBGPServerConfig)
```
taktv6, line 58: If we start using a ServerConfig structure shouldn't that contain the routerID and maybe even the listening addresses too?

BarbarossaTM: Fine with me, will refactor

taktv6 (approved):

PR #409, opened 2023-01-13
Add plumbing for GRP - gRPC Routing Protocol

BarbarossaTM:
This adds the first bits and pieces for a gRPC-based Routing Protocol (GRP). It will allow to leverage Bio-Routing as a proxy between classical routing protocols, such as BGP, IS-IS, and OSPF, and a user-defined gRPC service.

This commit contains the new GRP path type definition, the integration into existing route and path structures, as well as stripped down version of the AdjRIBIn and AdjRIBOut routing tables.

This assumes that #410 is merged :)


protocols/grp/adjRIBIn/adj_rib_in.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/bio-routing/bio-rd/net"
+	"github.com/bio-routing/bio-rd/route"
+	"github.com/bio-routing/bio-rd/routingtable"
+	"github.com/bio-routing/bio-rd/routingtable/filter"
+	"github.com/bio-routing/bio-rd/routingtable/vrf"
+	"github.com/bio-routing/bio-rd/util/log"
+)
+
+// AdjRIBIn represents an Adjacency RIB In as described in RFC4271
+type AdjRIBIn struct {
```
taktv6, line 15: This looks like a 1:1 copy of the adjRIBIn we have already. What is stiopping us from using the other?

protocols/grp/adjRIBOut/adj_rib_out.go
```diff
@@ -0,0 +1,245 @@
...
+	bnet "github.com/bio-routing/bio-rd/net"
+	"github.com/bio-routing/bio-rd/route"
+	"github.com/bio-routing/bio-rd/routingtable"
+	"github.com/bio-routing/bio-rd/routingtable/filter"
+	"github.com/bio-routing/bio-rd/routingtable/locRIB"
+	"github.com/bio-routing/bio-rd/util/log"
+)
+
+// AdjRIBOut represents an Adjacency RIB Out
+type AdjRIBOut struct {
```
taktv6, line 16: Same here.

protocols/grp/api/grp.proto
```diff
@@ -0,0 +1,20 @@
...
+
+package bio.bgp;
+
+import "net/api/net.proto";
+import "route/api/route.proto";
+
+option go_package = "github.com/bio-routing/bio-rd/protocols/grpcrp/api";
+
+message GRPRoute {
+    string vrf = 1;
```
taktv6, line 11: Given the rib in / rib out design: How does that map to VRFs? Are we going to have a RIB in and a RIB out per VRF then?

taktv6: If we think about GRP being kind of a BGP without BGP maybe this is then more a GRMP (gRPC Routing Monitoring Protocol)?

protocols/grp/adjRIBIn/adj_rib_in.go
BarbarossaTM: It's not, it's heavily stripped down, and any BGP specific bit is removed as GRP intentionally does not now about ASN and BGP Path attributes. That's also the reason why we can't just use the existing one.

protocols/grp/adjRIBOut/adj_rib_out.go
BarbarossaTM: Same as above :-D

protocols/grp/api/grp.proto
BarbarossaTM: GRP is designed to be VRF-aware, so every prefix is send/received with a VRF name provided, so we're basically multiplexing all VRFs over the gRPC connection and the client has to MUX/DEMUX this into the existing VRFs and AFs within Bio.

To GRP vs. GRMP: GRP is designed to be two way, so not only monitoring, hence I'd not add monitoring to the name, as it would be misleading.

BarbarossaTM:
Closing this in favor of smaller PRs

PR #411, opened 2023-01-14
Add gRPC-base Routing Protocol (GRP) path type

BarbarossaTM:
This adds the first bits and pieces for a gRPC-based Routing Protocol (GRP).  It will allow to leverage Bio-Routing as a proxy between classical routing protocols, such as BGP, IS-IS, and OSPF, and a user-defined gRPC service.

This commit contains the new GRP path type definition and the integration into existing route and path structures.

It's split from #409 to go in smaller steps.


protocols/grp/api/grp.proto
```diff
@@ -0,0 +1,18 @@
...
+option go_package = "github.com/bio-routing/bio-rd/protocols/grpcrp/api";
+
+message GRPRoute {
+    string vrf = 1;
+    bio.route.Route route = 2;
+    bool announcement = 3;
+}
+
+service GRPService {
+    rpc Sync(stream GRPRoute) returns (stream GRPRoute) {}
```
taktv6, line 17: Rename to RunSession

taktv6 (approved):

PR #417, opened 2023-01-19
BMP Receiver: Close Connection Explicitly and with log message

margau:
It was not clear why a connection was not properly accepted after the connection came from the wrong source.
This PR changes the behaviour by explicitly closing the BMP TCP Connection and adding a log output.


taktv6:
Looks good. Thanks for your contribution!

PR #418, opened 2023-01-19
ISIS: Extend integration testing

taktv6:
Extending ISIS integration tests. Mocking time here.


tests/isis_integration_test.go
```diff
@@ -59,4 +121,199 @@ func TestServer(t *testing.T) {
...
+		132,              // Type
+		4,                // Length
+		169, 254, 100, 0, // IP Address
+		// Area Addresses TLV
+		1,       // Type
+		3,       // Length
+		2,       // Area length
+		0x49, 0, // Area
+	}, pkt) {
+		return
```
BarbarossaTM, line 215: Why just `return` and not erroring out?

taktv6: assert.Equal logs what is wrong where. Thus just exiting here.
Checking the following assertions doesn't make sense here if the previous failed because they depend on the previous ones. if assertion A fails it implies all the following will most probably fail and the outputs of those failed asserts would just pollute the tests output.

BarbarossaTM (approved):
LGTM

PR #422, opened 2023-01-29
ISIS: Ensure adjacency goes down when interface does, trigger LSP gen on link up/down + refresh LSP when it gets old

taktv6:


BarbarossaTM (approved):

BarbarossaTM (approved):
LGTM

PR #424, opened 2023-02-01
ISIS: Minor improvements

taktv6:


net/ip.go
```diff
@@ -9,13 +9,37 @@ import (
...
+)
+
 // IP represents an IPv4 or IPv6 address
 type IP struct {
 	higher   uint64
 	lower    uint64
 	isLegacy bool
 }
 
+func (ip IP) IsLoopback() bool {
```
BarbarossaTM, line 23: While you're at it I'd include IPv6 link-local here too (and rename the method to something meaningful)

BarbarossaTM (approved):

PR #427, opened 2023-02-03
ObserveRIB: Add Support for observing a not yet ready RIB

margau:
When Observing a RIB, it might be useful to already start receiving the routes when an EOR marker was not yet received.
This PR adds an flag allowing to skip the readiness check to the GRPC API. The default behaviour is unchanged.


cmd/ris/risserver/server.go
```diff
@@ -192,9 +192,10 @@ func (s *Server) ObserveRIB(req *pb.ObserveRIBRequest, stream pb.RoutingInformat
 	if err != nil {
 		return status.New(codes.Unavailable, wrapGetRIBErr(err, req.Router, vrfID, ipVersion).Error()).Err()
 	}
-
```
BarbarossaTM, line 195: Nit: please keep the empty line in between the blocks here :-)

BarbarossaTM (commented):
This is neat, thanks!

I feel it would be nicer if we call the flag (and internal varaibles `allow_unready_rib` /  `allowUnready` to clearly indicate what the purpose this, otherwise I love this :)

margau:
Fixed the naming, and additionally added the flag enabled to ris mirror, and to risclient data types.

taktv6:
Thanks, Marvin!

BarbarossaTM:
Takt was faster :-D Thanks, indeed!

PR #429, opened 2023-02-08
Fix VRF creation in ris-mirror

margau:
VRFs for RIS-Mirror were not created with the correct name, this MR fixes that.


BarbarossaTM (approved):
Good catch, thanks for the fix!

PR #430, opened 2023-02-13
ris-mirror: Add API for removing one or all vrf ris clients at runtime

margau:
When using the ris-mirror API, it was only possible to add new VRFs, but not remove vrfs from the ris-mirror at runtime.
This leads to problems when for example an vrf/router is not available any longer, but cannot be removed from the ris-mirror, leading to failing observeRIB calls.
This PR adds basic support for gracefully removing an router/VRF at runtime.

### DIFF
--- a/cmd/ris-mirror/rismirror/rismirror.go
+++ b/cmd/ris-mirror/rismirror/rismirror.go
@@ -1,12 +1,14 @@
 package rismirror
 
 import (
+	"fmt"
 	"net"
 	"sync"
 
 	"github.com/bio-routing/bio-rd/cmd/ris-mirror/rismirror/metrics"
 	"github.com/bio-routing/bio-rd/protocols/bgp/server"
 	"github.com/bio-routing/bio-rd/routingtable/vrf"
+	"github.com/bio-routing/bio-rd/util/log"
 	"google.golang.org/grpc"
 )
 
@@ -33,6 +35,37 @@ func (rism *RISMirror) AddTarget(rtrName string, address net.IP, vrfRD uint64, s
 
 	r := rism.routers[rtrName].(*Router)
 	r.addVRF(vrfRD, sources)
+}
+
+func (rism *RISMirror) RemoveTarget(rtrName string, vrfRD uint64) error {
+	rism.routersMu.Lock()
+	defer rism.routersMu.Unlock()
+
+	if _, exists := rism.routers[rtrName]; !exists {
+		return fmt.Errorf("Router %s does not exist, cannot remove cleanly", rtrName)
+	}
+	r := rism.routers[rtrName].(*Router)
+	err := r.removeVRF(vrfRD)
+	if err != nil {
+		return err
+	}
+	if len(r.GetVRFs()) == 0 {
+		delete(rism.routers, rtrName)
+	}
+	return nil
+}
+
+func (rism *RISMirror) RemoveRouter(rtrName string) {
+	rism.routersMu.Lock()
+	defer rism.routersMu.Unlock()
+
+	if _, exists := rism.routers[rtrName]; !exists {
+		log.Errorf("Router %v does not exist, cannot remove cleanly", rtrName)
+		return
+	}
+	r := rism.routers[rtrName].(*Router)
+	r.dropAllVRFs()
+	delete(rism.routers, rtrName)
 }
 
 // GetRouter gets a router
--- a/cmd/ris-mirror/rismirror/router.go
+++ b/cmd/ris-mirror/rismirror/router.go
@@ -1,11 +1,13 @@
 package rismirror
 
 import (
+	"fmt"
 	"net"
 	"sync"
 
 	"github.com/bio-routing/bio-rd/risclient"
 	"github.com/bio-routing/bio-rd/routingtable/vrf"
+	"github.com/bio-routing/bio-rd/util/log"
 	"google.golang.org/grpc"
 
 	"github.com/bio-routing/bio-rd/cmd/ris/api"
@@ -74,15 +76,53 @@ func (r *Router) addVRF(rd uint64, sources []*grpc.ClientConn) {
 	defer r.vrfsMu.Unlock()
 
 	v := r.vrfRegistry.CreateVRFIfNotExists(vrf.RouteDistinguisherHumanReadable(rd), rd)
-	r.vrfs[rd] = newVRFWithMergedLocRIBs(v.IPv4UnicastRIB(), v.IPv6UnicastRIB())
+	r.vrfs[rd] = newVRFWithMergedLocRIBs(v.IPv4UnicastRIB(), v.IPv6UnicastRIB(), v)
 
 	for _, src := range sources {
-		r.connectVRF(rd, src, 4)
-		r.connectVRF(rd, src, 6)
+		rc4 := r._connectVRF(rd, src, 4)
+		rc6 := r._connectVRF(rd, src, 6)
+		// we cannot call v directly here, because the ris clients are only in the specific implementation of the vrf struct.
+		r.vrfs[rd].Clients = append(r.vrfs[rd].Clients, rc4, rc6)
 	}
 }
 
-func (r *Router) connectVRF(rd uint64, src *grpc.ClientConn, afi uint8) {
+func (r *Router) removeVRF(rd uint64) error {
+	r.vrfsMu.Lock()
+	defer r.vrfsMu.Unlock()
+	if r.GetVRF(rd) == nil {
+		return fmt.Errorf("VRF %v in Router %sv does not exist, cannot remove cleanly", rd, r.address.String())
+	}
+	// for removal, first stop all vrfs
+	r._stopVRF(rd)
+	// then we clean the map entries
+	vrf := r.vrfs[rd]
+	r.vrfRegistry.UnregisterVRF(vrf.vrf)
+	delete(r.vrfs, rd)
+	return nil
+}
+
+func (r *Router) dropAllVRFs() {
+	for _, vrf := range r.GetVRFs() {
+		r.removeVRF(vrf.RD())
+	}
+}
+
+func (r *Router) _stopVRF(rd uint64) {
+	// check if we even have an vrf for this rd
+	if _, exists := r.vrfs[rd]; !exists {
+		log.Errorf("Trying to stop invalid vrf %v", rd)
+		return
+	}
+	v := r.vrfs[rd]
+	// first step: stop all ris clients
+	for _, rc := range v.Clients {
+		rc.Stop()
+	}
+	// now we can clear the vrfClient list
+	r.vrfs[rd].Clients = make([]*risclient.RISClient, 0)
+}
+
+func (r *Router) _connectVRF(rd uint64, src *grpc.ClientConn, afi uint8) *risclient.RISClient {
 	rc := risclient.New(&risclient.Request{
 		Router:          r.name,
 		VRFRD:           rd,
@@ -91,6 +131,7 @@ func (r *Router) connectVRF(rd uint64, src *grpc.ClientConn, afi uint8) {
 	}, src, r.vrfs[rd].getRIB(afi))
 
 	rc.Start()
+	return rc
 }
 
 func apiAFI(afi uint8) api.ObserveRIBRequest_AFISAFI {
--- a/cmd/ris-mirror/rismirror/vrf.go
+++ b/cmd/ris-mirror/rismirror/vrf.go
@@ -1,6 +1,7 @@
 package rismirror
 
 import (
+	"github.com/bio-routing/bio-rd/risclient"
 	"github.com/bio-routing/bio-rd/routingtable/locRIB"
 	"github.com/bio-routing/bio-rd/routingtable/mergedlocrib"
 	"github.com/bio-routing/bio-rd/routingtable/vrf"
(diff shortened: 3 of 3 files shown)


cmd/ris-mirror/rismirror/rismirror.go
```diff
@@ -35,6 +36,41 @@ func (rism *RISMirror) AddTarget(rtrName string, address net.IP, vrfRD uint64, s
...
+	rism.routersMu.Lock()
+	defer rism.routersMu.Unlock()
+
+	if _, exists := rism.routers[rtrName]; !exists {
+		log.Errorf("Router %v does not exist, cannot remove cleanly", rtrName)
+		delete(rism.routers, rtrName)
+		return
+	}
+	r := rism.routers[rtrName].(*Router)
+	// remove chosen vrf
```
taktv6, line 49: I think this comment is pretty redundant with the next line.

cmd/ris-mirror/rismirror/rismirror.go
```diff
@@ -35,6 +36,41 @@ func (rism *RISMirror) AddTarget(rtrName string, address net.IP, vrfRD uint64, s
...
+	r.removeVRF(vrfRD)
+	// if no vrf is left, remove router
```
taktv6, line 51: Same here.

cmd/ris-mirror/rismirror/rismirror.go
```diff
@@ -35,6 +36,41 @@ func (rism *RISMirror) AddTarget(rtrName string, address net.IP, vrfRD uint64, s
...
+	rism.routersMu.Lock()
+	defer rism.routersMu.Unlock()
+
+	if _, exists := rism.routers[rtrName]; !exists {
+		log.Errorf("Router %v does not exist, cannot remove cleanly", rtrName)
+		delete(rism.routers, rtrName)
+		return
+	}
+	r := rism.routers[rtrName].(*Router)
+	// remove all vrfs
```
taktv6, line 67: Please move this to a method `dropAllVRFs()` on the `Router` type.

cmd/ris-mirror/rismirror/router.go
```diff
@@ -54,15 +56,38 @@ func (r *Router) GetVRFs() []*vrf.VRF {
 func (r *Router) addVRF(rd uint64, sources []*grpc.ClientConn) {
 	v := r.vrfRegistry.CreateVRFIfNotExists(vrf.RouteDistinguisherHumanReadable(rd), rd)
 
-	r.vrfs[rd] = newVRF(v.IPv4UnicastRIB(), v.IPv6UnicastRIB())
+	r.vrfs[rd] = newVRF(v)
 
+	// cleanup old ris clients of the vrf
+	r.stopVRF(rd)
```
taktv6, line 62: Why do we need to stop a freshly created VRF?

cmd/ris-mirror/rismirror/router.go
```diff
@@ -16,6 +16,7 @@ type Router struct {
 	address     net.IP
 	vrfRegistry *vrf.VRFRegistry
 	vrfs        map[uint64]*_vrf
+	vrfClients  map[uint64][]*risclient.RISClient
```
taktv6, line 19: Given this changes at runtime and is accessed from multiple go routines, access needs to be synchronized or am I missing anything here? The same applies to the `vrfs` member I guess.

taktv6 (changes requested):

cmd/ris-mirror/rismirror/router.go
margau: I've added a mutex here

cmd/ris-mirror/rismirror/rismirror.go
margau: Done.

cmd/ris-mirror/rismirror/rismirror.go
margau: Done.

cmd/ris-mirror/rismirror/router.go
taktv6: Now we have a deadlock here. `vrfClientsMu` gets locked here and also in `stopVRF()`.

cmd/ris-mirror/rismirror/router.go
```diff
@@ -12,10 +13,12 @@ import (
...
 // Router represents a router
 type Router struct {
-	name        string
-	address     net.IP
-	vrfRegistry *vrf.VRFRegistry
-	vrfs        map[uint64]*_vrf
+	name         string
+	address      net.IP
+	vrfRegistry  *vrf.VRFRegistry
+	vrfs         map[uint64]*_vrf
```
taktv6, line 19: I think access to `vrfs` also needs to be synchronized or am I missing something?

cmd/ris-mirror/rismirror/router.go
```diff
@@ -52,17 +56,51 @@ func (r *Router) GetVRFs() []*vrf.VRF {
...
+func (r *Router) stopVRF(rd uint64) {
+	r.vrfClientsMu.Lock()
+	defer r.vrfClientsMu.Unlock()
+
+	// first step: stop all ris clients
+	for _, rc := range r.vrfClients[rd] {
+		rc.Stop()
 	}
+	// now we can clear the vrfClient list
+	r.vrfClients[rd] = r.vrfClients[rd][:0]
```
taktv6, line 100: Please use `make([]*risclient.RISClient, 0)` here.

cmd/ris-mirror/rismirror/router.go
taktv6: Acquiring the lock once and then doing all the work is the right thing here. We need a `stopVRF()` function that doesn't lock on its own. Create `_stopVRF()` that does the same without the lock. Change `stopVRF()` to only do locking and call `_stopVRF()`.

cmd/ris-mirror/rismirror/router.go
```diff
@@ -12,10 +13,12 @@ import (
...
 type Router struct {
-	name        string
-	address     net.IP
-	vrfRegistry *vrf.VRFRegistry
-	vrfs        map[uint64]*_vrf
+	name         string
+	address      net.IP
+	vrfRegistry  *vrf.VRFRegistry
+	vrfs         map[uint64]*_vrf
+	vrfClients   map[uint64][]*risclient.RISClient
```
taktv6, line 20: I'm wondering why we don't maintain a VRFs clients within the `_vrf` object? Would be more straight forward I guess.

margau: Done. I was a bit careful about that, because of the relation to `vrf.VRF`.

cmd/ris-mirror/rismirror/router.go
margau: Moved the mutex to vrf access, as the clients are now inside the vrfs.

margau:
Hi @taktv6, do you have an update on this PR?

cmd/ris-mirror/rismirror/rismirror.go
```diff
@@ -35,6 +36,36 @@ func (rism *RISMirror) AddTarget(rtrName string, address net.IP, vrfRD uint64, s
...
+		return
+	}
+	r := rism.routers[rtrName].(*Router)
+	r.removeVRF(vrfRD)
+	if len(r.GetVRFs()) < 1 {
+		delete(rism.routers, rtrName)
+	}
+}
+
+func (rism *RISMirror) RemoveTargetFull(rtrName string) {
```
taktv6, line 55: Rename: RemoveRouter

cmd/ris-mirror/rismirror/router.go
```diff
@@ -52,17 +55,62 @@ func (r *Router) GetVRFs() []*vrf.VRF {
...
+	delete(r.vrfs, rd)
+}
+
+func (r *Router) dropAllVRFs() {
+	for _, vrf := range r.GetVRFs() {
+		r.removeVRF(vrf.RD())
+	}
+}
+
+func (r *Router) stopVRF(rd uint64) {
```
taktv6, line 93: Rename to _stopVRF and drop the comment. In the bio code base the underscore prefix indicates that locks must be hold when calling the method.

cmd/ris-mirror/rismirror/router.go
```diff
@@ -52,17 +55,62 @@ func (r *Router) GetVRFs() []*vrf.VRF {
...
+	// no lock here, because this method may be only called with locked vrfMu!
+
+	// check if we even have clients for this rd
+	if _, exists := r.vrfs[rd]; !exists {
```
taktv6, line 97: I think this check is not necessary. The for loop below doesn't do anything when the entry does not exist.

margau: The for loop cycles over ` r.vrfs[rd].Clients` - so we get a null pointer if `r.vrfs[rd]` is nil i'd say.

taktv6: No. Iterating over a nil slice is fine. It just doesn't iterate.
https://play.golang.com/p/SKpZkFaPdAx

taktv6: Oh, now I see. Why not something like this then:
```
v := r.vrfs[rd]
if v == nil {
   return
}

for _, rc := range v.Clients {
...
```

cmd/ris-mirror/rismirror/router.go
```diff
@@ -52,17 +55,60 @@ func (r *Router) GetVRFs() []*vrf.VRF {
...
+
+func (r *Router) _stopVRF(rd uint64) {
+	// check if we even have an vrf for this rd
+	v := r.vrfs[rd]
+	if v == nil {
+		return
+	}
+	// first step: stop all ris clients
+	for _, rc := range v.Clients {
+		if rc != nil {
```
taktv6, line 101: I'm wondering how a nil client would end up in the slice?

taktv6: Oh yes, if we ask to stop a VRF we should maybe know if it already exists.
Let me quickly check where we write to this slice. It should not contain nil at all.

margau: On a second thought, `func (rism *RISMirror) RemoveTarget(rtrName string, vrfRD uint64) {` would make even more sense, possibly with adjusting this method to return an error (instead of just logging it). Thoughts about this?

cmd/ris-mirror/rismirror/vrf.go
```diff
@@ -1,19 +1,24 @@
...
 
-func newVRF(locRIBIPv4Unicast, locRIBIPv6Unicast *locRIB.LocRIB) *_vrf {
+func newVRF(vrf *vrf.VRF) *_vrf {
 	return &_vrf{
-		ipv4Unicast: mergedlocrib.New(locRIBIPv4Unicast),
-		ipv6Unicast: mergedlocrib.New(locRIBIPv6Unicast),
+		vrf:         vrf,
+		ipv4Unicast: mergedlocrib.New(vrf.IPv4UnicastRIB()),
+		ipv6Unicast: mergedlocrib.New(vrf.IPv6UnicastRIB()),
+		Clients:     make([]*risclient.RISClient, 2),
```
taktv6, line 21: Here is the issue! This creates to nil entries in the slice. I guess you want to set capacity, not length.
`make([]*risclient.RISClient, 0, 2)`

cmd/ris-mirror/rismirror/router.go
taktv6: Yes. No error necessary here. The nil check can be removed after fixing the make (see my other comment below).

cmd/ris-mirror/rismirror/vrf.go
margau: Updated it

cmd/ris-mirror/rismirror/rismirror.go
```diff
@@ -35,6 +36,38 @@ func (rism *RISMirror) AddTarget(rtrName string, address net.IP, vrfRD uint64, s
...
+func (rism *RISMirror) RemoveTarget(rtrName string, vrfRD uint64) {
+	rism.routersMu.Lock()
+	defer rism.routersMu.Unlock()
+
+	if _, exists := rism.routers[rtrName]; !exists {
+		log.Errorf("Router %s does not exist, cannot remove cleanly", rtrName)
+		return
+	}
+	r := rism.routers[rtrName].(*Router)
+	if r.GetVRF(vrfRD) == nil {
```
taktv6, line 48: Why not make `removeVRF` return an `error` when the VRF does not exist?

cmd/ris-mirror/rismirror/rismirror.go
```diff
@@ -35,6 +36,38 @@ func (rism *RISMirror) AddTarget(rtrName string, address net.IP, vrfRD uint64, s
...
+		log.Errorf("VRF %v in Router %v does not exist, cannot remove cleanly", vrfRD)
+		return
+	}
+	r.removeVRF(vrfRD)
+	if len(r.GetVRFs()) < 1 {
```
taktv6, line 53: I guess we actually care if we have zero, so lets check `== 0`.

cmd/ris-mirror/rismirror/rismirror.go
margau: Sounds good, I've added some more error handling.

cmd/ris-mirror/rismirror/rismirror.go
margau: Adjusted that.

taktv6 (approved):
LGTM

PR #431, opened 2023-02-14
riscli: Add "get routers" feature

margau:
This PR allows riscli users to get all available routers with their vrfs.

### DIFF
--- a/cmd/riscli/routers.go
+++ b/cmd/riscli/routers.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	pb "github.com/bio-routing/bio-rd/cmd/ris/api"
+	"github.com/bio-routing/bio-rd/routingtable/vrf"
+	"github.com/bio-routing/bio-rd/util/log"
+	"github.com/urfave/cli"
+)
+
+// NewLPMCommand creates a new LPM command
+func NewGetRoutersCommand() cli.Command {
+	cmd := cli.Command{
+		Name:  "routers",
+		Usage: "get all routers and vrfs available",
+		Flags: []cli.Flag{},
+	}
+
+	cmd.Action = func(c *cli.Context) error {
+		conn := SetupGRPCClient(c)
+		defer conn.Close()
+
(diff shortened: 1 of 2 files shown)


cmd/riscli/routers.go
```diff
@@ -0,0 +1,48 @@
...
+	return cmd
+}
+
+func getRouters(c pb.RoutingInformationServiceClient, routerName string, vrfID uint64, vrf string) error {
+	resp, err := c.GetRouters(context.Background(), &pb.GetRoutersRequest{})
+	if err != nil {
+		return fmt.Errorf("unable to get client: %w", err)
+	}
+	for _, r := range resp.GetRouters() {
+		fmt.Printf("Router %v at %v\n", r.SysName, r.Address)
```
taktv6, line 44: Nit: In this case know we're having strings, right?
If yes, please change to `%s`.

cmd/riscli/routers.go
```diff
@@ -0,0 +1,48 @@
...
+		fmt.Printf("VRFs: %v\n\n", r.VrfIds)
```
taktv6, line 45: Why not print human readable route distinguisher? See https://pkg.go.dev/github.com/bio-routing/bio-rd@v0.1.7/routingtable/vrf#RouteDistinguisherHumanReadable

cmd/riscli/routers.go
```diff
@@ -0,0 +1,48 @@
...
+
+	cmd.Action = func(c *cli.Context) error {
+		conn := SetupGRPCClient(c)
+		defer conn.Close()
+
+		client := pb.NewRoutingInformationServiceClient(conn)
+		err := getRouters(client, c.GlobalString("router"), c.GlobalUint64("vrf_id"), c.GlobalString("vrf"))
+		if err != nil {
+			log.Errorf("Get Routers failed: %v", err)
+			os.Exit(1)
```
taktv6, line 29: Nit: Why not `log.Panicf()` instead of two error and exit? (Yes, I did that in the past...)

taktv6 (approved):

PR #433, opened 2023-02-23
feat: Implement exponential backoff for risclient

margau:
This is a first suggestion on how to implement an exponential backoff for the ris client,
avoiding the ris getting flooded in ObserveRIB retries.
It has been not yet been tested, and is intended as an RFC.


risclient/risclient.go
```diff
@@ -80,26 +83,43 @@ func (r *RISClient) stopped() bool {
...
+	orc, err := risc.ObserveRIB(context.Background(), r.req.toProtoRequest(), grpc.WaitForReady(true))
+	if err != nil {
+		log.WithError(err).Error("ObserveRIB call failed")
+		return err
+	}
 
+	err = r.serviceLoop(orc)
+	if err != nil {
 		r.serviceLoopLogging(err)
+	} else {
```
taktv6, line 104: Let's get rid of this else by returning err inside the if above.

risclient/risclient.go
```diff
@@ -23,6 +25,7 @@ type RISClient struct {
 	req    *Request
 	cc     *grpc.ClientConn
 	c      Client
+	b      *backoff.ExponentialBackOff
```
taktv6, line 28: I see we're having a bunch of very short names here. However, I think just be is not very descriptive. Let's rename it to backoff.

taktv6 (changes requested):
LGTM in general. Please address the 2 comments and we're good to go!

taktv6 (approved):

PR #434, opened 2023-02-26
RIS-Mirror: Refactoring

taktv6:
It was unclear in the RIS mirror which member of the `Router` object is authoritative regarding the VRFs.
The idea is to make the local map authoritative on existing VRFs. The `vrf.VRFRegistry` is only used to pass it to the metrics package to provide metrics.

### DIFF
--- a/cmd/ris-mirror/rismirror/router.go
+++ b/cmd/ris-mirror/rismirror/router.go
@@ -2,6 +2,7 @@ package rismirror
 
 import (
 	"net"
+	"sync"
 
 	"github.com/bio-routing/bio-rd/risclient"
 	"github.com/bio-routing/bio-rd/routingtable/vrf"
@@ -14,16 +15,17 @@ import (
 type Router struct {
 	name        string
 	address     net.IP
-	vrfRegistry *vrf.VRFRegistry
-	vrfs        map[uint64]*_vrf
+	vrfs        map[uint64]*vrfWithMergedLocRIBs // this is the authoritative data store for VRFs
+	vrfsMu      sync.RWMutex
+	vrfRegistry *vrf.VRFRegistry // this is only there so that the metrics functionality of the vrf package can be used
 }
 
 func newRouter(name string, address net.IP) *Router {
 	return &Router{
 		name:        name,
 		address:     address,
+		vrfs:        make(map[uint64]*vrfWithMergedLocRIBs),
 		vrfRegistry: vrf.NewVRFRegistry(),
-		vrfs:        make(map[uint64]*_vrf),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cmd/ris-mirror/rismirror/router.go
```diff
@@ -43,18 +45,36 @@ func (r *Router) Ready(vrf uint64, afi uint16) (bool, error) {
 
 // GetVRF gets a VRF by its RD
 func (r *Router) GetVRF(rd uint64) *vrf.VRF {
-	return r.vrfRegistry.GetVRFByName(vrf.RouteDistinguisherHumanReadable(rd))
+	r.vrfsMu.RLock()
+	defer r.vrfsMu.RUnlock()
+
+	_vrf := r.vrfs[rd]
+	if _vrf == nil {
+		return nil
```
BarbarossaTM, line 53: Shouldn't `return r.vrfs[rd]` suffice here?

cmd/ris-mirror/rismirror/router.go
```diff
@@ -14,16 +15,17 @@ import (
 type Router struct {
 	name        string
 	address     net.IP
-	vrfRegistry *vrf.VRFRegistry
-	vrfs        map[uint64]*_vrf
+	vrfs        map[uint64]*vrfWithMergedLocRIBs // this is the authoritative data store for VRFs
+	vrfsMu      sync.RWMutex
+	vrfRegistry *vrf.VRFRegistry // this is only there so that the metrics functionality of the vrf package can be used
```
BarbarossaTM, line 20: I think the comment is not true, given the 

```
r.vrfRegistry.CreateVRFIfNotExists(vrf.RouteDistinguisherHumanReadable(rd), rd)
```

below :-D

cmd/ris-mirror/rismirror/vrf.go
```diff
@@ -3,21 +3,23 @@ package rismirror
 import (
 	"github.com/bio-routing/bio-rd/routingtable/locRIB"
 	"github.com/bio-routing/bio-rd/routingtable/mergedlocrib"
+	"github.com/bio-routing/bio-rd/routingtable/vrf"
 )
 
-type _vrf struct {
+type vrfWithMergedLocRIBs struct {
+	vrf         *vrf.VRF
```
BarbarossaTM, line 10: This field seems to be unused?

cmd/ris-mirror/rismirror/router.go
taktv6: The point is when you remove/add VRFs you have to do it on the `vrfs` member first, holding the mutex. `vrfRegistry` is only there so that metrics of the VRFs can be exposed. It not authoritative.

cmd/ris-mirror/rismirror/vrf.go
taktv6: Not yet. It should be in a following PR by Marvin.

BarbarossaTM (approved):
LGTM

PR #449, opened 2023-08-31
Enforce non-empty ASPath on eBGP sessions and set local-pref 100 if none is set

BarbarossaTM:
This does two things:

**Hide routes received via eBGP when the AS-Path is empty**

I was wondering if it would be a good idea to just create an AS-Path with the peer ASN in it, but depending on what the peer did to their policies this may not reflect reality and may actually further support leaking prefixes or something. So I decided to err on the side of caution here and consider prefixes received via eBGP and w/o an AS-Path as invalid/hidden.
We _may_ want to ignore this check when learning prefixes via BMP, however I would hope that no BGPd out there would accept it.

**Default local-pref for eBGP prefixes to 100**

This is a follow-up on a previous PR, which only took effect when running BIO as BGP server and not as library. This fixes the residual places.


czerwonk (approved):
LGTM

PR #459, opened 2024-01-02
Lower the bar on building Bio-Routing

BarbarossaTM:
This is intended as the first step to lower the bar of entry on using Bio-Routing :-)

* Add a `Makefile` to build Bio-Routing commands and examples & test the code
* Update GitHub workflows to use make targets
* Update README to point out how to build and run `bio-rd`
* Some clean-ups

A good next step would be to start populating `Documentation/user` a bit more on more example configurations and how to use `bio-rd` and `bio-rdc` together.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
         run: .github/check-gofmt.sh
 
       - name: Test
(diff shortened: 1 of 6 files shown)


ricky-charlet:
Thanks. These would be *very* helpful to me and others approaching this project from scratch.

taktv6 (approved):

PR #464, opened 2024-04-15
Enable SO_REUSEPORT

netixx:
Enable SO_REUSEPORT on the TCP socket,

This enables multiple instances of the BGP server to be started on the same machine, binding to the same TCP port.

The kernel will load-balance the incoming connections to all listening instances.
Instances that have the correct BGP peers configured will take the connection,
and those who don't will reject it, letting another instance have a try when the router restarts the connection.

### DIFF
--- a/net/tcp/listen.go
+++ b/net/tcp/listen.go
@@ -9,7 +9,7 @@ import (
 )
 
 type ListenerFactoryI interface {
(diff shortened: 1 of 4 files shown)


taktv6:
Thanks for contributing. LGTM in general. But I think we should make this configurable to avoid confusion for those who accidentally start multiple instances they didn't intent to.

taktv6 (approved):
LGTM

PR #467, opened 2024-04-18
Allow setting MaxPaths, EcmpOnly and BestOnly for ObserveRIB request

netixx:
Solves github.com/bio-routing/bio-rd/issues/466 for the ris server part.


netixx:
Something looks wrong with the codecov action atm...

cmd/ris/api/ris.proto
```diff
@@ -58,6 +65,7 @@ message ObserveRIBRequest {
     }
     AFISAFI afisafi = 3;
     bool allow_unready_rib = 5;
+    optional ObserveRIBOptions observe_options = 6; // additional options for the observe request
```
taktv6, line 68: Is there a reason you're just not extending the `ObserveRIBRequest` instead of defining another message?

PR #470, opened 2024-05-10
fix: deadlock in ris-mirror vrf removal

tobikris:
The same Mutex was tried to Lock twice on removal of a VRF.


tobikris:
codecov "Rate limit reached."

taktv6 (approved):
LGTM

PR #471, opened 2024-07-01
fix: set AddPathLen for BGP Routes when received via grpc

Istvan91:
When using ris-mirror the routes weren't sorted by BestPath:

The ASPathLen value was not set when decoding BGP Paths from proto and defaulted back to zero => all routes where treated as having the same ASPathLen, regardless of the actual ASPath.


Istvan91:
fixed the corresponding test

taktv6 (approved):
LGTM

PR #474, opened 2024-07-21
Fix CI by dumping codecov and upgrade go

taktv6:
Codecov is heavily rate limited these days. For an org like bio it would cost us 5$ per month and user. Gotta look for an alternative. Dropping until then.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.20.x]
+        go-version: [1.22.x]
         platform: [ubuntu-latest]
     runs-on: ${{ matrix.platform }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -22,14 +22,5 @@ jobs:
...
         run: make test-coverage
 
-      # Upload Coverage Report
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v1
-        with:
-          file: ./coverage.txt
-          flags: unittests
-          name: codecov-${{ matrix.platform }}-${{ matrix.go-version }}
-          fail_ci_if_error: true
```
BarbarossaTM, line 32: How 'bout we set this to `false` and keep it for the time being? :thinking:

PR #487, opened 2025-03-19
Add behaviour to the get request in RIS

maxiestudies:
Adds functionality to fetch information from all routers if the router parameter in the grpc call is empty, without modifying the protobuf generated files


hikhvar:
Rebased in #496

Issue #488, opened 2025-03-20
isValidIdentifier() doesn't conform to rfc6286


bartdorlandt:
**Describe the bug**
The BGP identifier can be any 32-bit address, which doesn't have to be an actual address as it was in a previous RFC than rfc6286. (previous rfc4271)

Because of that a BGP identifier of "0.0.0.1" is now considered `false`, where it should be allowed, therefore true.

**Steps to Reproduce**
Set up a BGP peer with a BGP identifier of 0.0.0.1

**Expected behavior**
It should error on any uint32 value. Therefore continue process the BGP messages

**Additional context**
Code suggestion:

Current code:
```
func validateOpen(msg *BGPOpen) error {
	if msg.Version != BGP4Version {
		return BGPError{
			ErrorCode:    OpenMessageError,
			ErrorSubCode: UnsupportedVersionNumber,
			ErrorStr:     fmt.Sprintf("Unsupported version number"),
		}
	}
	if !isValidIdentifier(msg.BGPIdentifier) {
		return BGPError{
			ErrorCode:    OpenMessageError,
			ErrorSubCode: BadBGPIdentifier,
			ErrorStr:     fmt.Sprintf("Invalid BGP identifier"),
		}
	}

	return nil
}

func isValidIdentifier(id uint32) bool {
	addr := net.IP(convert.Uint32Byte(id))
	if addr.IsLoopback() {
		log.Errorf("BGP identifier is loopback address: %v", addr)
		return false
	}

	if addr.IsMulticast() {
		log.Errorf("BGP identifier is multicast address: %v", addr)
		return false
	}

	if addr[0] == 0 {
		log.Errorf("BGP identifier is network address: %v", addr)
		return false
	}

	if addr[0] == 255 && addr[1] == 255 && addr[2] == 255 && addr[3] == 255 {
		log.Errorf("BGP identifier is broadcast address: %v", addr)
		return false
	}

	return true
}
```

New code:
```
func validateOpen(msg *BGPOpen) error {
	if msg.Version != BGP4Version {
		return BGPError{
			ErrorCode:    OpenMessageError,
			ErrorSubCode: UnsupportedVersionNumber,
			ErrorStr:     fmt.Sprintf("Unsupported version number"),
		}
	}

	return nil
}

```

PR #489, opened 2025-03-20
isValidIdentifier() doesn't conform to rfc6286

bartdorlandt:
Fixes [#488](https://github.com/bio-routing/bio-rd/issues/488)

### DIFF
--- a/protocols/bgp/packet/decoder.go
+++ b/protocols/bgp/packet/decoder.go
@@ -3,14 +3,12 @@ package packet
 import (
 	"bytes"
 	"fmt"
-	"net"
 
 	"github.com/bio-routing/bio-rd/util/decode"
-	"github.com/bio-routing/tflow2/convert"
 )
 
 const (
-	addPathTupleSize = 4
+	addPathTupleSize         = 4
 	extendedNextHopTupleSize = 6
 )
 
@@ -357,13 +355,6 @@ func validateOpen(msg *BGPOpen) error {
 			ErrorStr:     "unsupported version number",
 		}
 	}
-	if !isValidIdentifier(msg.BGPIdentifier) {
-		return BGPError{
-			ErrorCode:    OpenMessageError,
-			ErrorSubCode: BadBGPIdentifier,
-			ErrorStr:     "invalid BGP identifier",
-		}
-	}
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


protocols/bgp/packet/decoder.go
```diff
@@ -393,27 +384,6 @@ func decodeExtendedNextHopCapability(buf *bytes.Buffer, capLength uint8) (Extend
...
-		return false
-	}
-
-	if addr.IsMulticast() {
-		return false
-	}
-
-	if addr[0] == 0 {
-		return false
-	}
```
BarbarossaTM, line 408: According to RFC6286, this part must not be removed:

> If the BGP Identifier field of the OPEN message is zero, or if it
> is the same as the BGP Identifier of the local BGP speaker and the
> message is from an internal peer, then the Error Subcode is set to
> "Bad BGP Identifier".

Ideally we'd also check the latter part here then.

BarbarossaTM (changes requested):
Thanks for your issue and PR!

To fully implement RFC6286 this seems to need some more changes to check the identifier collision for iBGP peers and, ideally, the collision handling as of section 2.3.

I'd prefer to have at least the first part conforming to RFC6286, ideally both. Do you want to take a shot at updating the PR to fully support RFC6286? :-)

protocols/bgp/packet/decoder.go
bartdorlandt: From what I read from the RFC, the whole address can't be zero (0.0.0.0). The frist portion can be. Therefor 0.0.0.1 is a valid BGP identifier.

See these sections:

>  [2.2](https://datatracker.ietf.org/doc/html/rfc6286#section-2.2).  Open Message Error Handling
> 
>    For a BGP speaker that supports the AS-wide Unique BGP Identifier,
>    the OPEN message error handling related to the BGP Identifier is
>    modified as follows:
> 
>       If the BGP Identifier field of the OPEN message is zero, or if it
>       is the same as the BGP Identifier of the local BGP speaker and the
>       message is from an internal peer, then the Error Subcode is set to
>       "Bad BGP Identifier".

And, not the 2nd paragraph of the first chapter. 

>  [1](https://datatracker.ietf.org/doc/html/rfc6286#section-1).  Introduction
>    Currently, the BGP Identifier of a BGP speaker is specified as a
>    valid IPv4 host address assigned to the BGP speaker [[RFC4271](https://datatracker.ietf.org/doc/html/rfc4271)].  In
>    addition, the deployed BGP code requires that two BGP speakers be of
>    distinct BGP Identifiers in order to establish a BGP connection.
> 
>    To accommodate situations where the current requirements for the BGP
>    Identifier are not met (such as in the case of an IPv6-only network),
>    this document relaxes the definition of the BGP Identifier to be a
>    4-octet, unsigned, non-zero integer and relaxes the "uniqueness"
>    requirement so that only AS-wide uniqueness of the BGP Identifiers is
>    required.  These revisions to the base BGP specification do not
>    introduce any backward compatibility issues.

BarbarossaTM: You're right, I should have looker a little closer and in the surrounding code, I actually changed myself :see_no_evil: 

Sorry, was a long day :grin:

BarbarossaTM (approved):

bartdorlandt:
Thank you. Do you have a timeline for this to reach a release?

PR #494, opened 2025-05-21
fix: avoid internal errors on query parsing for ris-lg

matthieugouel:
Hey 👋 

In ris-lg, if the client is sending an origin ASN or AFI which cannot be converted from string to int with Atoi, it receives a 500 internal error. I believe It would be more meaningful to serve a 400 Bad Request instead.

### DIFF
--- a/cmd/ris-lg/lg/lg.go
+++ b/cmd/ris-lg/lg/lg.go
@@ -166,13 +166,13 @@ func (l *LookingGlass) Routes(w http.ResponseWriter, r *http.Request) {
 
 	originASN, err := getOriginASN(r)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


BarbarossaTM (approved):
Indeed, thanks!

matthieugouel:
The CI is not passing because of linting errors currently present on the main branch. 
I tried to resolve it but it's widespread in the codebase, so I think it's better to address those errors on a separate PR, what do you think @BarbarossaTM ?

BarbarossaTM:
> The CI is not passing because of linting errors currently present on the main branch. I tried to resolve it but it's widespread in the codebase, so I think it's better to address those errors on a separate PR, what do you think @BarbarossaTM ?

Agree, I opened #495 to fix those. Once that's merged, I'll rebase your PR and merge it.

PR #500, opened 2026-03-29
fix: prevent nil dereference in Path.ECMP() and BGPPath.Prepend()

joshuafuller:
Hi! We're using bio-rd as an embedded BGP speaker in a project. We ran into a crash in testing that we think may be a bug in `Path.ECMP()` — wanted to share our findings and a proposed fix.

### How we found it

We have a scenario where two nodes each originate the same anycast prefix (`10.0.99.0/24`) as a `StaticPathType` route via `LocRIB.AddPath()`. When the two nodes peer over eBGP, each receives the other's copy of that prefix as a `BGPPathType` route. At that point the locRIB has two paths for the same prefix — one static, one BGP — and the process panics:

```
panic: runtime error: invalid memory address or nil pointer dereference

goroutine 144 [running]:
github.com/bio-routing/bio-rd/route.(*BGPPath).ECMP(...)
    route/bgp_path.go:229
github.com/bio-routing/bio-rd/route.(*Path).ECMP(...)
    route/path.go:70
github.com/bio-routing/bio-rd/route.(*Route).updateEqualPathCount(...)
    route/route.go:318
github.com/bio-routing/bio-rd/route.(*Route).PathSelection(...)
    route/route.go:210
github.com/bio-routing/bio-rd/routingtable/locRIB.(*LocRIB).AddPath(...)
```

Since this happens inside the FSM goroutine, we can't `recover()` from it — it kills the process.

### What we think is happening

`Path.ECMP()` (`route/path.go:67-78`) switches on `p.Type` and calls the type-specific ECMP method on both `p` and `q`, but doesn't check that `q` has the same type first:

```go
func (p *Path) ECMP(q *Path) bool {
    switch p.Type {
    case BGPPathType:
        return p.BGPPath.ECMP(q.BGPPath)  // q.BGPPath is nil if q is StaticPathType
```

We noticed that `Path.Select()` in the same file already has the guard:

```go
func (p *Path) Select(q *Path) int8 {
    if p.Type > q.Type {
        return 1
    }
    if p.Type < q.Type {
        return -1
    }
    // ... switch on p.Type ...
```

So it looks like `ECMP()` may have been missed when that guard was added to `Select()`.

We also noticed that `AdjRIBOut.redistributePath()` explicitly handles redistribution from `StaticPathType` to BGP, which suggests mixed path types in the same RIB is an intended use case.

### Second issue: BGPPath.Prepend() with nil ASPath

While investigating, we also hit a related panic in `BGPPath.Prepend()` (`route/bgp_path.go:655`):

```go
if len(*b.ASPath) == 0 {  // panics if b.ASPath is nil
```

This happens when a `BGPPath` is constructed without initializing `ASPath` — `Prepend()` is then called during the export pipeline and dereferences the nil pointer. A nil check before the existing length check would make this defensive.

### What this PR does

- `Path.ECMP()`: adds `if p.Type != q.Type { return false }` before the switch (mirrors the existing guard in `Select()`)
- `BGPPath.Prepend()`: initializes `ASPath` via `types.NewASPath(nil)` if nil before the existing `len()` check
- 5 test cases covering both issues

All existing tests pass. gofmt, go vet, and gocyclo (under 15) are clean. Happy to adjust the approach if you'd prefer a different fix. Thanks for maintaining bio-rd!

### DIFF
--- a/route/bgp_path.go
+++ b/route/bgp_path.go
@@ -652,6 +652,10 @@ func (b *BGPPath) Prepend(asn uint32, times uint16) {
 		return
 	}
 
+	if b.ASPath == nil {
+		b.ASPath = types.NewASPath(nil)
+	}
+
 	if len(*b.ASPath) == 0 {
 		b.insertNewASSequence()
 	}
--- a/route/mixed_path_type_test.go
+++ b/route/mixed_path_type_test.go
@@ -1,0 +1,173 @@
+package route
+
+import (
+	"testing"
+
+	bnet "github.com/bio-routing/bio-rd/net"
+	"github.com/bio-routing/bio-rd/protocols/bgp/types"
+)
+
+// TestECMP_MixedPathTypes_BGPvsStatic verifies that ECMP does not panic
+// when comparing a BGPPathType path against a StaticPathType path.
+// This is the primary crash: bio-rd calls p.BGPPath.ECMP(q.BGPPath)
+// without checking that q is also BGPPathType. When q is StaticPathType,
+// q.BGPPath is nil → nil pointer dereference.
+func TestECMP_MixedPathTypes_BGPvsStatic(t *testing.T) {
+	bgpPath := &Path{
+		Type: BGPPathType,
+		BGPPath: &BGPPath{
+			BGPPathA: &BGPPathA{
+				NextHop:   bnet.IPv4(1).Ptr(),
+				LocalPref: 100,
+				Origin:    0,
+				Source:    bnet.IPv4(1).Ptr(),
+			},
+			ASPath:    types.NewASPath([]uint32{64512}),
+			ASPathLen: 1,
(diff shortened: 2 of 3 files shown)


joshuafuller:
Updated the PR to address the review comments:

- `TestRouteUpdateEqualPathCount_MixedTypes`: now calls `r.PathSelection()` before checking `ECMPPathCount()`, and asserts count == 1
- `TestBGPPathPrepend_NilASPath`: added assertions for `ASPathLen`, segment count, and that the prepended ASN (64512) is in the first segment
- `BGPPath.Prepend()` nil init: switched to `types.NewASPath(nil)` for consistency with existing helpers

BarbarossaTM:
Thanks for fix! I'll have a deeper look later (and try to disable copilot :cry: )

